PR #1, opened 2018-08-09
Feature cloud optics

RobertPincus:


RobertPincus:
Adding module and data for cloud optics calculation. Data and original code provided by Mike Iancono at AER.

PR #18, opened 2019-01-11
OpenACC Fixes

mrnorman:
read the commit messages.

### DIFF
--- a/build/Makefile.conf.olcf_pgi
+++ b/build/Makefile.conf.olcf_pgi
@@ -8,7 +8,7 @@ export FC = mpif90
 #
 # Optimized
 #
-export FCFLAGS  = -O0 -g -ta=nvidia,cc70,ptxinfo -Minfo=accel -ta=tesla:autocompare
-export FCFLAGS  = -O0 -g -ta=nvidia,cc70,ptxinfo -Minfo=accel -ta=tesla:autocompare
+export FCFLAGS  = -O3 -g -ta=nvidia,cc70,ptxinfo -Minfo=accel
+export FCFLAGS  = -O3 -g -ta=nvidia,cc70,ptxinfo -Minfo=accel
 #export FCFLAGS  = -m64 -O3 -traceback -assume realloc_lhs -extend-source 132 -qopt-report-phase=vec
 #export F77FLAGS = -m64 -O3 -traceback                                        -qopt-report-phase=vec
--- a/build/Makefile.conf.olcf_pgi
+++ b/build/Makefile.conf.olcf_pgi
@@ -1,0 +1,14 @@
+#
+# PGI Fortran on Ascent/Summit
+#
+# Fortran compiler command
+export FC = mpif90
+
+# Fortran compiler flags
+#
+# Optimized
+#
+export FCFLAGS  = -O0 -g -ta=nvidia,cc70,ptxinfo -Minfo=accel -ta=tesla:autocompare
+export FCFLAGS  = -O0 -g -ta=nvidia,cc70,ptxinfo -Minfo=accel -ta=tesla:autocompare
+#export FCFLAGS  = -m64 -O3 -traceback -assume realloc_lhs -extend-source 132 -qopt-report-phase=vec
+#export F77FLAGS = -m64 -O3 -traceback                                        -qopt-report-phase=vec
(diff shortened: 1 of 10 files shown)


RobertPincus:
Matt, Ben - 

You know, I think we ought to keep the OpenACC changes on their own branch until we have a version working end-to-end. I'd like to keep master for codes that can be deployed and putting these changes on the develop branch means they'll get merged to master sooner rather than later. Would you please open a PR to merge against my openacc-kernels branch instead? 

- Robert

mrnorman:
Done!

RobertPincus:
Thanks, Matt. 

Can I confirm the state of the code? My understanding is that his implementation builds under PGI and Gnu but doesn’t run under Gnu. Everything runs on the GPU but the data movement isn’t optimized yet so it’s not particularly efficient. We still haven’t figured out how to structure the data statements so the spectrally-depdenent data lives on the GPU from start to end. Is that about right?

RobertPincus:
Thanks, Matt. 

Can I confirm the state of the code? My understanding is that his implementation builds under PGI and Gnu but doesn’t run under Gnu. Everything runs on the GPU but the data movement isn’t optimized yet so it’s not particularly efficient. We still haven’t figured out how to structure the data statements so the spectrally-depdenent data lives on the GPU from start to end. Is that about right?

RobertPincus:
not "his implementation" but "this implementation".

mrnorman:
Yes, that's right. It will be slower than the CPU version as it is. But it runs on the GPU and gives the right answer for now, and this allows us to test GPU alongside CPU to know when the codes diverge at all (so changes can be more easily propagated to the kernels-openacc code when necessary).

As we found out this past Friday, it's going to be rough going optimizing data "all at once" because apparently, the GPU address for some of the arrays is changing from one place in the code to another inexplicably. So I think the easiest approach is to start aggregating data movement from the "bottom up" starting in the lowest level routines and working our way up the callstack from there, but I'm open to ideas.

vectorflux:
Robert, Matt, 

Just checking in here, since there is an ICON developers' meeting Feb. 11-15, and I am supposed to mention the 'status of ICON on GPU'..  Of course I will also check with Gustavo for his take of how the RRTMGP experimentation is going.  I'm not sure he is involved in any additional OpenACC porting.

It is apparent the GPU implementation still has some distance to go on this end.  Could you folks give an appraisal on whether there is bandwidth available to finish up the port?   I've been holding back in the hope that someone more capable will pick up the ball.  But it is possible that I might free up some time, or that we can find someone else at CSCS to help out.  

Let me know what is going on, so I can add mention in the meeting.

mrnorman:
Hi "vectorflux" (I'm not sure who you are by name),

What's needed is to optimize the data movement and most likely port some additional parts of the workflow to GPUs in the SW and LW calcs. My approach to this is typically to put "parent" data statements around kernels at the lowest level and to build up from there. This is the most incremental approach and the easiest to debug as you go.

I will definitely not get to this by mid February because I have other deadlines I'm on the hook for. However, I expect to get this done in late February or early March because it's required for our E3SM INCITE runs at OLCF. The timing depends on whether we consider RRTMGP or an improvement to the super-parameterization port to be the priority for our project.

vectorflux:
Sorry, I use this pseudonym to hide my real identity...

That is good news that the OpenACC code satisfies a milestone for INCITE.
I'll get Gustavo's take on things in Hamburg Feb. 11-15.  It is possible
that he is also working on the OpenACC port himself.

Thanks for the quick response.

--Will

mrnorman:
Haha, hey Will! That's a great superhero name.

No problem. It's taken a lot longer than it should, but I'm hoping things will go smoothly once I can get to it. I remember us doing this at the hackathon and having nearly everything on the GPU.

mrnorman:
Oh, nevermind again, github cuts your name off when I look at it on the webpage but it showed up in the e-mail.

RobertPincus:
@vectorflux My understanding is that the MPI (Gustavo) is focusing on implementing RTE+RRTMGP within ICON and coding the parts of the problem that don't concern the solvers or the gas optics (e.g. cloud and aerosol optics, surface properties, etc.) We (me, @mrnorman and @brhillman  from the DOE, hopefully you and CSCS are on deck to complete the GPU port of the kernels. 

I have been without access to a GPU host though I expect to get access to the DOE's Summit in the coming days. I would very much like to set aside some time with Matt and Ben to woodshed this project to completion. I have some ideas about how to approach the final problem though I'll need some help to actually get it done.

PR #19, opened 2019-01-29
Consistent use of `logical(wl)` throughout code

Chiil:
This is the fix that makes the code use `c_bool` wherever logicals are used (`wl` kind as defined in `mo_rte_kind.F90`. I have done a full implementation throughout the code, but *not* in the extensions, as I have no methods of testing those.

In order to make the code run on GNU and Intel compilers, I had to convert the usage of `this%get_ngas()` in `dimension` definitions into `get_ngas(this)`.

To let clean builds work on single thread make, I had to swap the order of compilation of `librrtmgp.a` and `librte.a` as the latter needs to be compiled first.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -8,7 +8,7 @@ VPATH = $(RTE_DIR):$(RTE_KERNEL_DIR):$(RRTMGP_DIR):$(RRTMGP_KERNEL_DIR)
 include Makefile.conf
 include Makefile.rules
 
-all: librrtmgp.a librte.a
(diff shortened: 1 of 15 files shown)


RobertPincus:
Hi Chiel - 

Thanks for this and sorry to have not tested against Gnu. Do you mind to change the PR to merge against the new branch hotfix-logical-types? I'd prefer to stage the changes rather than push them directly onto master. I also want to check performance before making the changes everywhere (as opposed to just arguments to kernels with C bindings.) 

- Robert

PR #23, opened 2019-05-22
Travis CI

RobertPincus:
Can we make Travis CI go?


clementval:
Now building and running with PGI 19.4, GNU 8 and GNU 9

PR #29, opened 2019-06-11
Remove nullify on declaration for thread safety

brhillman:
Remove nullify statements on declaration of pointers in subroutines to ensure
thread safety. When pointers get assigned in
declarations, they implicitly get a save attribute and are assumed static. This
is a problem when it occurs in a threaded region, so this code was NOT
thread-safe before. Removing the `=> NULL()` does not change the behavior of the
code for non-threaded applications, but does ensure thread-safety. Tested with the RFMIP clear-sky example, and confirmed bit-for-bit.

### DIFF
--- a/rrtmgp/mo_gas_optics_rrtmgp.F90
+++ b/rrtmgp/mo_gas_optics_rrtmgp.F90
@@ -390,7 +390,7 @@ contains
     integer :: igas, idx_h2o ! index of some gases
     ! Number of molecules per cm^2
     real(wp), dimension(ncol,nlay), target  :: col_dry_arr
-    real(wp), dimension(:,:),       pointer :: col_dry_wk => NULL()
+    real(wp), dimension(:,:),       pointer :: col_dry_wk
     !
     ! Interpolation variables used in major gas but not elsewhere, so don't need exporting
     !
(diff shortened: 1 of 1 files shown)


RobertPincus:
Hi Ben, thanks a ton, you are the best. A few questions: 
1) Can you please revise this to merge against branch "develop"? 
2) The other place that pointers are initialized to NULL() is in ty_fluxes_broadband and its descendants.  Here the initialization seems important since we test whether the pointer is associated to know if people will want a certain calculation done. Are there circumstances where this could cause a problem? 

- Robert

brhillman:
@RobertPincus I agree the initialization is important in the `ty_fluxes_*` definitions, and if initialization is removed then users would need to be cautious to make sure that the unused fluxes were nullified manually. I'm not sure the implementation as it stands would cause a problem though, since those are in type definitions and not procedures. I'm definitely not an OpenMP expert (Pat Worley is responsible for finding this bug fix, by the way). I have confirmed that leaving them as-is still permits threaded builds of E3SM to run without error, so they might be fine to leave as-is. I vote we merge this PR as I have it here with just the fixes in the subroutine declarations, and if we can find an example later that breaks the fluxes then we can revisit how to deal with that problem.

brhillman:
Awesome, thanks Robert!

PR #36, opened 2019-08-06
Config for PGI 19.4 on Daint

phmarti:
PGI 19 switched the default backend to LLVM. In order to avoid linking errors, newer version of HDF5 and NetCDF need to be used as they were built with PGI 19.1.


RobertPincus:
Phillipe, thanks very much for this, can you change the merge target to branch `develop`?

phmarti:
It should now be cleanly based on develop

PR #39, opened 2019-08-08
Make minimum value k depend on floating point precision

Chiil:
Make the minimum value of k in the two-stream kernels relative to the precision of the floating point numbers. This makes the solver work in single precision with sufficient accuracy.

### DIFF
--- a/rte/kernels/mo_rte_solver_kernels.F90
+++ b/rte/kernels/mo_rte_solver_kernels.F90
@@ -498,6 +498,9 @@ contains
     real(wp) :: RT_term(ncol)
     real(wp) :: exp_minusktau(ncol), exp_minus2ktau(ncol)
 
+    ! Minimum value is defined at 1e4 larger than epsilon.
+    real(wp), parameter :: k_min = 1.e4_wp * epsilon(1._wp)
+
     real(wp), parameter :: LW_diff_sec = 1.66  ! 1./cos(diffusivity angle)
     ! ---------------------------------
     do j = 1, nlay
@@ -518,7 +521,7 @@ contains
       !   of < 0.1% in Rdif down to tau = 10^-9
       k(1:ncol) = sqrt(max((gamma1(1:ncol,j) - gamma2(1:ncol,j)) * &
                            (gamma1(1:ncol,j) + gamma2(1:ncol,j)),  &
-                           1.e-12_wp))
+                           k_min))
       exp_minusktau(1:ncol) = exp(-tau(1:ncol,j)*k(1:ncol))
 
       !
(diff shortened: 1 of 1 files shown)


RobertPincus:
Cheil, thanks for this suggestion. As I said I'm going to think about this - it's empirically ok but it doesn't converge properly, nor is the threshold set with respect to the root cause of the problem. I might  try to use this PR to motivate a more proper solution.

peterukk:
I think it's a good idea, at least as a temporary solution, so people can run the code in single precision. I tested the fix also with ecRAD and it works also there - as expected, but good to confirm with another dataset. 

What is the root cause of the problem, incidentally, if not k?

peterukk:
Robin Hogan and I spent some time looking at this issue with inaccuracies in shortwave reflectance-transmittance computations in single precision. 

I don't remember all the ins and outs of it anymore but the conclusion was that combining an adjusted threshold value for minimum K,  the existing RTE formulation with the "merge" to protect the division in RT_term (which wasn't in ecRAD before), and an added security to ensure Tdir <= 1 - Tnoscat - Rdir (credit to Robin), seems to make the results secure and accurate, based on extensive offline validation and also some IFS simulations. For the RFMIP profiles, the max differences in fluxes relative to a double precision result reduce to 2-3 * e-02 W/m2. These changes  are in a development branch of ecRAD and improves performance quite a bit from the previous IFS solution which was to do the reflectance-transmittance computations in double precision

See [the modified sw_two_stream in my fork](https://github.com/peterukk/rte-rrtmgp-nn/blob/main/rte/kernels/mo_rte_solver_kernels.F90#L1271)

RobertPincus:
Thanks for the excellent contribution @peterukk. I'm a bit buried with teaching but will merge some version of this in a few weeks.

Chiil:
This is great. We have also a set of fixes for our SP version, I will check with Menno where there is overlap and compare against these.

MennoVeerman:
I noticed another problem in single precision, in the interpolation routine this time. When concentrations of some gases are very low, the division `gas_1/(gas_1+ratio_eta_half*gas_2) (=eta)` seems to becomes exactly one instead of slight lower than one as in double precision. Consequently, `mod(loceta, 1), where loceta=eta*float(neta-1)` returns 0 in SP en ~1 in DP, leading to very different interpolation fractions.
However, I am note sure whether this is an issue that requires (immediate) fixing, because so far it only happened with ozone concentrations below 1e-4 ppb, which is not exactly realistic anyway

RobertPincus:
Thanks, @MennoVeerman This is a good catch.

PR #44, opened 2019-09-26
Nvidia opt

alexeedm:


RobertPincus:
@alexeedm This looks great. Can you change the target of the PR to branch `develop`?

rrtmgp/kernels/mo_reorder_kernels.F90
```diff
@@ -24,18 +24,34 @@ subroutine reorder_123x312_kernel(d1, d2, d3, array_in, array_out) &
     real(wp), dimension(d1, d2, d3), intent( in) :: array_in
     real(wp), dimension(d3, d1, d2), intent(out) :: array_out
 
-    integer :: i1, i2, i3
+    integer :: i1, i2, i3, i10, i30, i1diff, i3diff
+    integer, parameter :: tile = 32
```
RobertPincus: This is a little mysterious and magic-numbery. Can you add comments to explain the reasoning?

rrtmgp/kernels/mo_reorder_kernels.F90
```diff
@@ -24,18 +24,34 @@ subroutine reorder_123x312_kernel(d1, d2, d3, array_in, array_out) &
...
 
-    !$acc parallel loop collapse(3) &
-    !$acc&     copyout(array_out(:d3,:d1,:d2)) &
-    !$acc&     copyin(array_in(:d1,:d2,:d3))
+
+    !$acc parallel vector_length(tile*tile) &
+    !$acc&     copyout(array_out) &
+    !$acc&     copyin(array_in)
+    !$acc loop gang collapse(3)
     do i2 = 1, d2
```
RobertPincus: The code is substantially less clear than a simple transposition. That's ok, but again please add a few lines of comments explaining that you're breaking the array up into pieces to do the transposition?

rrtmgp/kernels/mo_reorder_kernels.F90
```diff
@@ -44,18 +60,35 @@ subroutine reorder_123x321_kernel(d1, d2, d3, array_in, array_out) &
...
-    !$acc&     copyout(array_out(:d3,:d2,:d1)) &
-    !$acc&     copyin(array_in(:d1,:d2,:d3))
-    do i1 = 1, d1
-      do i2 = 1, d2
-        do i3 = 1, d3
-          array_out(i3,i2,i1) = array_in(i1,i2,i3)
+    integer :: i1, i2, i3, i10, i30, i1diff, i3diff, idiff
+    integer, parameter :: tile = 32
+
+    !$acc parallel vector_length(tile*tile) &
```
RobertPincus: As above - please comment to explain what's going on.

rte/mo_util_array.F90
```diff
@@ -34,73 +34,46 @@ module mo_util_array
...
-    end do
-    any_vals_less_than_1D = (minValue < minVal)
-#else
-    any_vals_less_than_1D = any(array < minVal)
-#endif
-end function any_vals_less_than_1D
-  !-------------------------------------------------------------------------------------------------
-  logical function any_vals_less_than_2D(array, minVal)
+    !$acc kernels copyin(array)
+    minValue = minval(array)
```
RobertPincus: Oooh, I hadn't realized you could support intrinsics like minval on the GPU. That's much cleaner.

alexeedm:
Just did, resolved a few conflicts. Let's see if the CI check passes

rte/kernels/mo_fluxes_broadband_kernels.F90
```diff
@@ -33,22 +33,21 @@ pure subroutine sum_broadband(ncol, nlev, ngpt, spectral_flux, broadband_flux) b
...
-      end do
-    end do
-    !$acc parallel loop collapse(3)
-    do igpt = 2, ngpt
-      do ilev = 1, nlev
-        do icol = 1, ncol
-          !$acc atomic update
-          broadband_flux(icol, ilev) = broadband_flux(icol, ilev) + spectral_flux(icol, ilev, igpt)
+
+        my_bb_flux = 0.0_wp
```
RobertPincus: Can you explain (to me, not necessarily in the code) why we use a scalar and only update the array element at the end?

rte/kernels-openacc/mo_rte_solver_kernels.F90
```diff
@@ -962,12 +973,12 @@ subroutine adding(ncol, nlay, ngpt, top_at_1, &
     real(wp), dimension(ncol,nlay+1,ngpt), intent(inout) :: flux_dn
     ! ------------------
     integer :: icol, ilev, igpt
-    real(wp), dimension(nlay+1) :: albedo, &  ! reflectivity to diffuse radiation below this level
+    real(wp), dimension(ncol,nlay+1,ngpt) :: albedo, &  ! reflectivity to diffuse radiation below this level
```
RobertPincus: My understanding from your comments in the Slack is that this is exposing a compiler bug, yes? If so, could you please add comments to that effect, explaining how we'd revert to a lower-memory option if that matters? 

Will you pursue this compiler fix?

rte/kernels-openacc/mo_rte_solver_kernels.F90
```diff
@@ -143,12 +129,37 @@ subroutine lw_solver_noscat(ncol, nlay, ngpt, top_at_1, D, weight,
...
+          ! Optical path and transmission, used in source function and transport calculations
+          !
+          tau_loc(icol,ilay,igpt) = tau(icol,ilay,igpt)*D(icol,igpt)
+          trans  (icol,ilay,igpt) = exp(-tau_loc(icol,ilay,igpt))
+          !
+          ! Weighting factor. Use 2nd order series expansion when rounding error (~tau^2)
+          !   is of order epsilon (smallest difference from 1. in working precision)
+          !   Thanks to Peter Blossey
+          !
+          fact = merge((1._wp - trans(icol,ilay,igpt))/tau_loc(icol,ilay,igpt) - trans(icol,ilay,igpt), &
```
RobertPincus: I had these two kernels separated for conceptual reasons. Is the performance gain quite large?

rrtmgp/kernels/mo_reorder_kernels.F90
alexeedm: This is basic blocking for more efficient cache usage, should also help on CPU. The threadblock needs to work on a small tile of the array such that non-contiguous reads (along the second or third array dimension) are effectively cached.
However, a better way to do the reordering is to use some CPU and GPU specific libraries.

rrtmgp/kernels/mo_reorder_kernels.F90
alexeedm: The two inner loops ensure that we're reading and writing the data in blocks of `tile*tile`. So the outer loops encode the blocks, while the inner loops go over all the elements in each of the blocks. That helps keeping the elements in such a block completely in the fastest cache.

rrtmgp/kernels/mo_reorder_kernels.F90
alexeedm: That's basically the same as I answered above

rrtmgp/kernels/mo_reorder_kernels.F90
RobertPincus: I realize this is what you're doing but since it's a pattern that might not be crystal-clear to naive readers, so I'm hoping you'll add comments to the code explaining.

rte/mo_util_array.F90
alexeedm: Yeah, I agree!

rrtmgp/kernels/mo_reorder_kernels.F90
RobertPincus: Oh, sorry - here I meant the specific choice of 32 as the tile size. The comments about the pattern apply to the code below.

rte/kernels/mo_fluxes_broadband_kernels.F90
alexeedm: The scalar in principle should reside in processor register (on CPU and GPU). This is the fastest memory available, and so it is useful since `my_bb_flux` is updated every iteration of the inner loop.
However, the compilers would likely do such an optimization anyway, I just wanted to make it totally explicit

rte/kernels-openacc/mo_rte_solver_kernels.F90
alexeedm: Right, that part is buggy. About the lower-memory option: the OpenACC compiler should implicitly transform the low-memory variant into what I wrote, because the GPU doesn't have enough per-thread storage. So for the GPU there should be no difference. If you think it is better to use low-memory variable for the CPU, I could easily change that (e.g., use an architecture-dependent macro)

I will file the compiler bug and follow it.

rte/kernels-openacc/mo_rte_solver_kernels.F90
alexeedm: I well understand your conceptual reasons. From the performance point of view, that's about 1ms less out of total ~30ms.
What I could suggest is to factor out the inner code into a routine, such that this routine has no loops, just the per-point stencil operations. Then we can put the routines together in the loop and the compiler will inline them.

rrtmgp/kernels/mo_reorder_kernels.F90
alexeedm: 32 just empirically the best choice for the GPU. I can definitely add comments here

RobertPincus:
@alexeedm I'll wait for another commit with changes that address these suggestions, yes?

alexeedm:
Yes, exactly, coming soon!

rte/kernels-openacc/mo_rte_solver_kernels.F90
RobertPincus: Thanks to the commit I see what you mean. Is the point here to save a kernel launch? If it's really 1 part in 30 I might opt for clarity. 

The point of separate loops was that variables `trans` and `tau_loc` are used in other parts of the code as well.

alexeedm: It's not just the kernel launch, it's memory reuse (`tau_loc` and `trans`). It is 1 part in 30 in the overall GPU runtime of the long-wave example, not just for this routine. The final decision is yours anyway!

alexeedm: Ideally you'd like you fuse as many kernels as possible in one big loop calling point-wise routines. That can save a lot of memory traffic and decrease the runtime quite a bit, but it would require a too much of code restructuring

rte/kernels-openacc/mo_rte_solver_kernels.F90
RobertPincus: Can you add comments to the code documenting this choice? 

The CPU code is distinct, no need to worry about that.

rte/kernels-openacc/mo_rte_solver_kernels.F90
RobertPincus: I think I can live with this... maybe you can just replace the routine, rather than making a `_stencil` version?

alexeedm: Sure, I went with `_stencil` simply because the other routine had a C interface and I thought you want it to be visible from outside. So I just stick to the original name, w/o `_stencil`, right?

RobertPincus: Ah, right. 

I have been thinking about revising the C-visible API so that only the top-level routines are visible (e.g. `lw_solver_noscat()`), but that will take a bit more time. 

So just leave things as they are.

rte/kernels-openacc/mo_rte_solver_kernels.F90
alexeedm: Done!

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -599,10 +672,12 @@ subroutine compute_Planck_source(                        &
...
     !
     !$acc parallel loop collapse(3)
-    do igpt = 1, ngpt
+    do icol = 1, ncol, 2
       do ilay = 1, nlay
-        do icol = 1, ncol
-          lay_src(igpt,ilay,icol) = pfrac(igpt,ilay,icol) * planck_function(gpoint_bands(igpt),ilay,icol)
+    do igpt = 1, ngpt
+          lay_src(igpt,ilay,icol  ) = pfrac(igpt,ilay,icol  ) * planck_function(gpoint_bands(igpt),ilay,icol)
+          if (icol <= ncol) &
```
RobertPincus: Shouldn't this test (line 679) be `icol < col`? But why is the test even needed?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -599,10 +672,12 @@ subroutine compute_Planck_source(                        &
     ! Map to g-points
     !
     !$acc parallel loop collapse(3)
-    do igpt = 1, ngpt
+    do icol = 1, ncol, 2
```
RobertPincus: It seems like you're explicitly unrolling a loop here - why?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -402,10 +402,81 @@ subroutine gas_optical_depths_minor(ncol,nlay,ngpt,        &
     integer  :: icol, ilay, iflav, igpt, imnr
     integer  :: gptS, gptE
     integer  :: minor_start, minor_loc, extent
+
+    real(wp),    dimension(ncol,nlay,ngpt) :: tau_t
```
RobertPincus: It looks like `tau_t` isn't ever used again?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -402,10 +402,81 @@ subroutine gas_optical_depths_minor(ncol,nlay,ngpt,        &
...
+    real(wp) :: myplay, mytlay, mycol_gas_h2o, mycol_gas_imnr, mycol_gas_0
+    real(wp) :: myfminor(2,2)
```
RobertPincus: Two things I'm confused about here. First, why are all these local variables helpful/necessary? Second, it looks like this new large block of code is for the case in which all the g-point intervals are the same (`all(minor_limits_gpt(2,:) - minor_limits_gpt(1,:) == gpt_diff`). This is true now but quite unlikely to be true generally. Is this a big performance win on the GPUs?

rte/kernels-openacc/mo_rte_solver_kernels.F90
```diff
@@ -413,109 +414,138 @@ end subroutine sw_solver_noscat
...
+    real(wp), dimension(ncol,nlay,ngpt) :: Rdif, Tdif, Rdir, Tdir, Tnoscat
+    real(wp), dimension(ncol,nlay,ngpt) :: source_up, source_dn
+    real(wp), dimension(ncol     ,ngpt) :: source_srf
+    ! ------------------------------------
+    !
+    ! Cell properties: transmittance and reflectance for direct and diffuse radiation
+    !
+    !$acc enter data copyin(tau, ssa, g, mu0, sfc_alb_dir, sfc_alb_dif, flux_dn, flux_dir)
+    !$acc enter data create(Rdif, Tdif, Rdir, Tdir, Tnoscat, source_up, source_dn, source_srf, flux_up)
+    call sw_two_stream(ncol, nlay, ngpt, mu0, &
```
RobertPincus: These two routines - `sw_two_stream` and `sw_source_2_str` would also be good candidates for the loop fusion/stencil treatment, no? Is there a reason you did this for the LW sources and not here? Is the computation/memory profile different?

rte/kernels-openacc/mo_rte_solver_kernels.F90
```diff
@@ -356,55 +357,55 @@ end subroutine lw_solver_2stream
...
-      else
-        ! layer index = level index
-        ! previous level is up (+1)
-        !$acc parallel loop collapse(2)
-        do igpt = 1, ngpt
-          do icol = 1, ncol
-            do ilev = nlay, 1, -1
-              flux_dir(icol,ilev,igpt) = flux_dir(icol,ilev+1,igpt) * exp(-tau(icol,ilev,igpt)*mu0_inv(icol))
-            end do
+      end do
```
RobertPincus: Github marks these lines as changed but I don't see differences. Do I miss something?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
alexeedm: Whoops. Indeed it has to be `icol < ncol`. That is because I unrolled the loop, and `icol+1` may exceed the array bounds if `ncol` is odd

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
alexeedm: Added comments. Increased number of memory transactions help to improve memory bandwidth quite a bit, and this loop is pretty expensive

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
alexeedm: Whoops! Removed

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
alexeedm: The local variables are needed to remove repeated memory reads inside the innermost (`do imnr = 1, extent`) loop. They help reduce the memory traffic quite a bit.
As for the performance, the routine `gas_optical_depths_minor` is the most expensive in long wave example. Unoptimized, it accounts for 60% of the runtime on Piz Daint P100 and 80% of the runtime on V100. Optimized version decreases the time of this routine by about 10x.

What would be the typical g-point intervals? Is it possible to keep them same, or it would be totally unreasonable? If they are not all the same, but pretty close, this optimization should also work, just slightly changed.

rte/kernels-openacc/mo_rte_solver_kernels.F90
alexeedm: They are indeed good candidates to be fused. The reason was that they have different loop pattern:  `sw_source_2_str` has layer dependency, and the inner loop has to be sequential. But it is still possible, and beneficial, to fuse these routines

rte/kernels-openacc/mo_rte_solver_kernels.F90
alexeedm: Maybe whitespace?..

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
RobertPincus: I'm a little surprised that the compiler doesn't save the result of the memory read. No? If not then we should implement this change in the alternative form of the loop as well. 

I don't think we can promise to keep all the g-point intervals the same. Does it help to know they always be in the range 1 to roughly 16? How narrow does the range of intervals need to be before the helps?

rte/kernels-openacc/mo_rte_solver_kernels.F90
RobertPincus: Let's come back to this.

alexeedm: I've implemented the change, and it indeed helps. It is just a bigger refactoring, I'd rather open a new pull request for that

RobertPincus: OK, that's great, but indeed let's do it as part of a different PR.

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
alexeedm: It is difficult to know exactly what the compiler would do. It is likely that it will use the registers, but it may think that the register pressure is already high and rely on cache instead.
Regarding the other variant I don't see many opportunities to store stuff in registers there because of a different data access pattern.

I think it is very helpful to know that the intervals are usually less than 16 (or 20, but something small anyway). Then we can always use the optimized variant, and just use the largest interval as `gpt_diff`. But I'll need to see the other typical cases to tell you more

RobertPincus: We don't have other variants for the moment. They will come in the form of new data but not for a while (few months). We can assume, for the moment, that the maximum number of g-points per band (`gpt_diff`) will not be larger than 32. 

If the code you wrote will be more efficient than the original for values of `gpt_diff` between 1 and 32 then I'd propose you code for variable number and omit the other branch. I don't mind having different kernels for the CPU and GPU versions but having two versions of the loops with the same kernel-y bits inside seems confusing. 

Would you mind?

alexeedm: Seems good to me, I'll make the changes now. For `gpt_diff < 32` the new version should definitely be faster. And later with more data I can come back and see if there is anything else I can do to optimize that routine.

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
RobertPincus: I don't quite understand this answer. Does it mean that you're explicitly telling the compiler that it can work on two sets of calculations at once? Isn't that obvious from the loop collapse?

alexeedm: On the GPU typically every loop iteration will be mapped to a single CUDA thread. Only a limited number of threads can be active at a time, just like in the CPU. What I did, is mapped two iteration from different memory locations onto single thread, so that each of the threads can issue two memory instructions. This allows more memory transactions in flight on the GPU, and therefore higher bandwidth (about 15-20% higher in that case). The general thing is called Instruction-Level Parallelism https://en.wikipedia.org/wiki/Instruction-level_parallelism

RobertPincus: Thanks, that's interesting. I'm only surprised that the compiler can't make (or be told to make) this kind of optimization without restructuring code.

PR #45, opened 2019-09-26
Changes for making the branch usable with icon's build system

m214089:


RobertPincus:
Thanks for these. Would be great to find a better solution in the long term where you were only compiling the files from extensions/ that you are using.

PR #47, opened 2019-09-27
fusing sw_two_stream and sw_source_2str

alexeedm:
Here is the second PR we were talking about. Fusing `sw_two_stream` and `sw_source_2str` is actually very beneficial, because you can get rid of the temporary arrays `Rdir`, `Tdir` and `Tnoscat`.

I've also changed the `_stencil` variants to work on scalars. That's a more proper way, I should have done it in the beginning. Makes it possible to call those subroutines with different indexing, and also helps compiler inline them.


RobertPincus:
Closed by 142317d

PR #48, opened 2019-09-28
GPU updates

RobertPincus:


RobertPincus:
We've moved on from this commit, so we'll merge later.

PR #50, opened 2019-10-02
Icon integration folder fix

antarcticrainforest:
We were running into issues with the icon build system as it was only copying kernel related contents from 
```rrtmgp/kernels-openacc``` and ```rte/kernels-openacc``` source folders. 

Yet for the source from these folders content from ```rte/kernels``` and ```rrtmgp/kernels``` have to be present. 

This small PR tries to fix this by moving the _common_ source (for cpu/openacc) version to  a common place and renaming the kernel directory to kernel-default.


RobertPincus:
This was resolved on the ICON side.

PR #56, opened 2019-10-31
CI on Piz Daint with azure pipelines

alexeedm:


RobertPincus:
We expect CI on this merge to fail because it uses the wrong project at CSCS.

PR #59, opened 2019-11-22
Add alternate angle

jdelamere:
Will merge into develop but there are conflicts to be resolved.  Also, still in the testing phase, but this pull request will allow code improvements before the final merge to develop/master.

Add-alternate-angle was created from master.


RobertPincus:
Jen, please remove the CSV file and Python script from the pull request as a first step.

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -1236,6 +1249,58 @@ function get_col_dry(vmr_h2o, plev, tlay, latitude) result(col_dry)
...
+    real(wp) :: trans_total
+
+    ncol = optical_props%get_ncol()
+    ngpt = optical_props%get_ngpt()
+    nbnd = optical_props%get_nband()
+
+    err_msg=""
+
+    ! Check match in dimensions between tau and optimal_angle_fit
+    if (nbnd .ne. size(this%optimal_angle_fit,2)) err_msg = &
```
RobertPincus: Better to check that the optical properties and gas optics class have the same spectral discretization using `gpoints_are_equal()` from `ty_optical_props`

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -1236,6 +1249,58 @@ function get_col_dry(vmr_h2o, plev, tlay, latitude) result(col_dry)
...
+      "gas_optics%compute_optimal_angles: optimal_angle_fit different dimension (nbnd)"
+    if (size(this%optimal_angle_fit,1) .ne. 2) err_msg = &
+      "gas_optics%compute_optimal_angles: optimal_angle_fit different dimension for linear fit (2)"
+
+    ! Check dimensions of incoming optimal_angles array
```
RobertPincus: Please merge in branch `develop` if you haven't already and use `extents_are()` from `mo_rte_util_array`

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -1236,6 +1249,58 @@ function get_col_dry(vmr_h2o, plev, tlay, latitude) result(col_dry)
...
+    ncol = optical_props%get_ncol()
+    ngpt = optical_props%get_ngpt()
+    nbnd = optical_props%get_nband()
+
+    err_msg=""
+
+    ! Check match in dimensions between tau and optimal_angle_fit
+    if (nbnd .ne. size(this%optimal_angle_fit,2)) err_msg = &
+      "gas_optics%compute_optimal_angles: optimal_angle_fit different dimension (nbnd)"
+    if (size(this%optimal_angle_fit,1) .ne. 2) err_msg = &
```
RobertPincus: It's not a bad idea to check the size of `this%optimal_angle_fit` but this should be done at assignment, not here.

rte/mo_rte_lw.F90
```diff
@@ -205,12 +211,22 @@ function rte_lw(optical_props, top_at_1, &
         !$acc enter data copyin(optical_props%tau)
         error_msg =  optical_props%validate()
         if(len_trim(error_msg) > 0) return
-        call lw_solver_noscat_GaussQuad(ncol, nlay, ngpt, logical(top_at_1, wl), &
-                              n_quad_angs, gauss_Ds(1:n_quad_angs,n_quad_angs), gauss_wts(1:n_quad_angs,n_quad_angs), &
+        if (present(lw_Ds)) then
+          lw_Ds_wt = gauss_wts(1,1)
```
RobertPincus: Why make this assignment rather than simply passing the value?

rte/mo_rte_lw.F90
```diff
@@ -70,16 +71,21 @@ function rte_lw(optical_props, top_at_1, &
...
+    real(wp), dimension(:,:),   &
+                      optional, intent(in   ) :: lw_Ds
     character(len=128)                        :: error_msg   ! If empty, calculation was successful
     ! --------------------------------
     !
     ! Local variables
     !
     integer :: ncol, nlay, ngpt, nband
     integer :: n_quad_angs
+    integer :: iquad
```
RobertPincus: Is this variable used?

rte/mo_rte_lw.F90
```diff
@@ -70,16 +71,21 @@ function rte_lw(optical_props, top_at_1, &
...
     integer :: icol, iband, igpt
     real(wp), dimension(:,:,:), allocatable :: gpt_flux_up, gpt_flux_dn
     real(wp), dimension(:,:),   allocatable :: sfc_emis_gpt
+    real(wp) :: lw_Ds_wt
```
RobertPincus: We should check that `n_quad_angle == 1` or isn't supplied if `lw_Ds` is supplied - they are mutually exclusive.

RobertPincus:
@jdelamere Is there a timeline for addressing these changes?

jdelamere:
Removed extra files from repo; however, new tests have been put in place for this new capability.  Please do not yet remove this branch even after merger unless that is a better idea than keeping it in place for continued work on this problem.

rte/mo_rte_lw.F90
jdelamere: My notes say to make a clear variable for passing it in BUT I can pass in directly if that is better.

rte/mo_rte_lw.F90
jdelamere: n_quad_angs defaults to one but is reassigned to the value of n_gauss_angles if that variable is present.  Where does it make sense to put this check?

RobertPincus:
@jdelamere I've made some changes to the implementation in this repo. Among other things I changed the name of the coefficients file back to it's original text, since this won't change answers. All the changes simple but it would be worth looking at on your end. 

Please update the testing and validation repo to provide a test of correctness. (This might just mean updating your branch of that repo.) Let me know when you can run the test/val repo against branch `add-alternate-angle` of this repo and I'll merge onto `develop.`

rte/mo_rte_lw.F90
RobertPincus: Changed to pass the value in directly

rrtmgp/mo_gas_optics_rrtmgp.F90
RobertPincus: Changed to use `extents_are()`

rte/mo_rte_lw.F90
RobertPincus: Didn't resolve - `n_gauss_angles` is ignored if `lw_Ds` are supplied

rrtmgp/mo_gas_optics_rrtmgp.F90
RobertPincus: Implemented this.

rrtmgp/mo_gas_optics_rrtmgp.F90
RobertPincus: Decided to leave this check out, since we don't do the same kinds of checks on any of the other data supplied from the coefficients file.

jdelamere:
This effort is being transitioned to add-alternate-angle-2 branch. Closing this pull request and opening a new one on that branch.

PR #60, opened 2019-12-18
implementation of the Tang approach

inpolonsky:
Robert,

Please merge tang3 to develop


rte/kernels-openacc/mo_rte_solver_kernels.F90
```diff
@@ -1197,5 +1198,310 @@ subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="apply_
...
+  end subroutine lw_solver_1rescl
+  ! -------------------------------------------------------------------------------------------------
+  !
+  ! LW transport, no scattering, multi-angle quadrature
+  !   Users provide a set of weights and quadrature angles
+  !   Routine sums over single-angle solutions for each sets of angles/weights
+  !
+  ! ---------------------------------------------------------------
+ 
+  subroutine lw_solver_1rescl_GaussQuad(ncol, nlay, ngpt, top_at_1, nmus, Ds, weights, &
```
RobertPincus: It's an interesting question - do we need the Gaussian quadrature version of this code? I'm not sure we'd ever use more than one angle.

rte/kernels-openacc/mo_rte_solver_kernels.F90
```diff
@@ -1197,5 +1198,310 @@ subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="apply_
       end do
     end if
   end subroutine apply_BC_0
+  ! ---------------------------------------------------------------
+  !
+  ! Tang
+  !
+  ! ---------------------------------------------------------------
+  subroutine lw_solver_1rescl(ncol, nlay, ngpt, top_at_1, D, weight,                             &
```
RobertPincus: Consistent a comment below this seems like a lot of repeated code. Could `lw_solver_1rescl()` be formulated to call `lw_solver_1str()` and then adjust the fluxes in a second step?

rte/mo_optical_props.F90
```diff
@@ -482,6 +537,22 @@ function copy_and_alloc_nstr(this, nmom, ncol, nlay, spectral_desc, name) result
     if(err_message /= "") return
     err_message = this%alloc_nstr(nmom, ncol, nlay)
   end function copy_and_alloc_nstr
+  ! --- Tang ------------------------------------------------------------------------
```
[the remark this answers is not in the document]
inpolonsky: fixed. changed to 1rescl in comments

rte/mo_optical_props.F90
```diff
@@ -1169,4 +1324,66 @@ function get_name(this)
...
+    real(wp) :: xx, ssal, yy
+    !$acc enter data copyin(tau, ssa, g)
+    !$acc enter data create(tauLoc, scaling)
+
+    !$acc parallel loop collapse(3)
+    do igpt=1,ngpt
+      do ilay=1,nlay
+        do icol=1,ncol
+          ssal = ssa(icol, ilay, igpt)
+          xx = ssal*(1._wp - g(icol, ilay, igpt)) / 2._wp
```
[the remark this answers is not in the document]
inpolonsky: fixed. xx->wb, yy->scaleDev

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,290 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+    do ilev = 1, nlay
+      radn_up(:,ilev+1) =  trans(:,ilev) * radn_up(:,ilev) +  source_up(:,ilev)
+      do icol=1,ncol
+         if ( ssa(icol,ilev) > 1e-6 )  then
+        !  
+        ! here ssa is used to store parameter wb/[(]1-w(1-b)] of Eq.21 of the Tang's paper
+        ! explanation of factor 0.4 note A of Table
+        !
+            xx = 0.4_wp*ssa(icol,ilev)*&
+                   ( radn_dn(icol,ilev+1)*(1.-trans(icol,ilev )**2 ) - &
```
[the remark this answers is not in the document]
inpolonsky: x**2 replaced with x*x.  Please check https://software.intel.com/en-us/forums/intel-fortran-compiler/topic/508466 saying that _"The compiler recognizes idioms such as **2 and **3 and generates the most efficient code for it. I recommend using the syntax that is clearest for the human reader"_

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,290 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+    ! Upward propagation
+    do ilev = 1, nlay
+      radn_up(:,ilev+1) =  trans(:,ilev) * radn_up(:,ilev) +  source_up(:,ilev)
+      do icol=1,ncol
+         if ( ssa(icol,ilev) > 1e-6 )  then
+        !  
+        ! here ssa is used to store parameter wb/[(]1-w(1-b)] of Eq.21 of the Tang's paper
+        ! explanation of factor 0.4 note A of Table
+        !
+            xx = 0.4_wp*ssa(icol,ilev)*&
```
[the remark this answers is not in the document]
inpolonsky: xx->adjustmentFactor

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,290 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+    !   radn_dn(:,ilev) = trans(:,ilev  )*radn_dn(:,ilev+1) + source_dn(:,ilev)
+    ! end do
+
+    ! Surface reflection and emission
+    ! radn_up(:,     1) = radn_dn(:,     1)*sfc_albedo(:) + source_sfc(:)
+    ! Upward propagation
+    do ilev = 1, nlay
+      radn_up(:,ilev+1) =  trans(:,ilev) * radn_up(:,ilev) +  source_up(:,ilev)
+      do icol=1,ncol
+         if ( ssa(icol,ilev) > 1e-6 )  then
```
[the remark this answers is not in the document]
inpolonsky: The test is performed to avoid computation of the adjustment factor for cloud free layers (they are usually outnumber the cloudy layer).

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,290 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+    ! do ilev = nlay, 1, -1
+    !   radn_dn(:,ilev) = trans(:,ilev  )*radn_dn(:,ilev+1) + source_dn(:,ilev)
+    ! end do
+
+    ! Surface reflection and emission
+    ! radn_up(:,     1) = radn_dn(:,     1)*sfc_albedo(:) + source_sfc(:)
+    ! Upward propagation
+    do ilev = 1, nlay
+      radn_up(:,ilev+1) =  trans(:,ilev) * radn_up(:,ilev) +  source_up(:,ilev)
+      do icol=1,ncol
```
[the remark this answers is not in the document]
inpolonsky: The adjustment factor for rad_up (rad_dn) for a given layer depends  on the rad_up (rad_dn) correspondingly that are not a layer parameter. Howerer, the adjustmant factor may be represented as 
adjustmentFactor_up(dn) = A_up(dn) radn_dn(up) + B_up(dn) source_dn(up)+C_dn(up) source_dn. Parameters A, B, C can be computed in lw_transport_1scl

rte/kernels-openacc/mo_rte_solver_kernels.F90
inpolonsky: 2 quadrature angles provides more accurate result than a single angle. So I would rather keep the capability.

rte/kernels-openacc/mo_rte_solver_kernels.F90
inpolonsky: The radiance fro the next level depends on the radiance on the previous level. So the adjustment to a radiance at given level has to be propagated to the followings levels. It seems to be simpler doing that way despite a lot of repeated code. As I suggested some  repeated code can be moved to lw_transport_1scl if representation for the adjustmant factor as 
adjustmentFactor = A_up(dn) radn_dn(up) + B_up(dn) source_dn(up)+C_dn(up)
is used

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -489,13 +489,13 @@ function compute_gas_taus(this,                       &
...
     !
     idx_h2o = string_loc_in_array('h2o', this%gas_names)
-    !$acc enter data create(col_dry_wk, col_dry_arr, col_gas)
     if (present(col_dry)) then
       col_dry_wk => col_dry
     else
       col_dry_arr = get_col_dry(vmr(:,:,idx_h2o), plev) ! dry air column amounts computation
       col_dry_wk => col_dry_arr
     end if
+    !$acc enter data create(col_dry_wk, col_dry_arr, col_gas)
```
RobertPincus: @inpolonsky Do you understand why this !$acc enter data create statements is after the pointer assignment in your branch and above the pointer assignment on develop? This seems like a place where the code should not be changing.

rte/mo_rte_lw.F90
```diff
@@ -56,7 +57,7 @@ module mo_rte_lw
   function rte_lw(optical_props, top_at_1, &
                   sources, sfc_emis,       &
                   fluxes,                  &
-                  inc_flux, n_gauss_angles) result(error_msg)
+                  inc_flux, n_gauss_angles, type2stream) result(error_msg)
```
RobertPincus: Since we have only two choices here - two-stream or re-scaled - I would much prefer a logical variable here, like `use_2str` or similar. (No error checking required!) If the variable is not supplied we assume we will use the re-scaling solution.

rte/mo_rte_lw.F90
```diff
@@ -225,6 +250,9 @@ function rte_lw(optical_props, top_at_1, &
                                sfc_emis_gpt, sources%sfc_source,       &
                                gpt_flux_up, gpt_flux_dn)
         !$acc exit data delete(optical_props%tau, optical_props%ssa, optical_props%g)
+        else
+           write(error_msg, "('lw_solver(...ty_optical_props_2str...) processing type canbe (0,1). entered: ', I2)") type2stream
```
RobertPincus: This error checking isn't needed if we use a logical variables - but also please note that we never call `write()` within the core of `rte` but instead return strings.

inpolonsky:
Robert:

Thanks. I have to be more careful with  !$acc  directives. I am fixing the code and about to push the changes.

With regard to write. FORTRAN uses write also to put information into character buffer. So I used it to fill the error_message string as I need to add an integer (type2stream) to it. Do you have any objection of such a usage?

Regards

Igor

Igor Polonsky
Atmospheric and Environmental Research, Inc.
131 Hartwell Avenue
Lexington, MA  02421-3126
office: 781-761-2228
fax: 781-761-2499
e-mail: ipolonsky@aer.com<mailto:ipolonsky@aer.com>



On 1/6/20 4:26 PM, Robert Pincus wrote:

@RobertPincus requested changes on this pull request.

A few changes to make, please.

________________________________

In rrtmgp/mo_gas_optics_rrtmgp.F90<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_RobertPincus_rte-2Drrtmgp_pull_60-23discussion-5Fr363489150&d=DwMCaQ&c=birp9sjcGzT9DCP3EIAtLA&r=PGXxSfQJAcousBe_GlGvaQ&m=gyhm3bPHjNJLTWfp9h5_SobTh2dkA1FVII3m-nxUauI&s=VQFx3WwAyxxJntJMvfRbkuPBs8SxaMSsQl4gdKBvCBk&e=>:

>      if (present(col_dry)) then
       col_dry_wk => col_dry
     else
       col_dry_arr = get_col_dry(vmr(:,:,idx_h2o), plev) ! dry air column amounts computation
       col_dry_wk => col_dry_arr
     end if
+    !$acc enter data create(col_dry_wk, col_dry_arr, col_gas)


@inpolonsky<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_inpolonsky&d=DwMCaQ&c=birp9sjcGzT9DCP3EIAtLA&r=PGXxSfQJAcousBe_GlGvaQ&m=gyhm3bPHjNJLTWfp9h5_SobTh2dkA1FVII3m-nxUauI&s=jMcdNI5taQx2HugnT6C0cimuVRUNXgDIvqdE11XXdtk&e=> Do you understand why this !$acc enter data create statements is after the pointer assignment in your branch and above the pointer assignment on develop? This seems like a place where the code should not be changing.

________________________________

In rte/mo_rte_lw.F90<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_RobertPincus_rte-2Drrtmgp_pull_60-23discussion-5Fr363489840&d=DwMCaQ&c=birp9sjcGzT9DCP3EIAtLA&r=PGXxSfQJAcousBe_GlGvaQ&m=gyhm3bPHjNJLTWfp9h5_SobTh2dkA1FVII3m-nxUauI&s=ZsyXMfvXtbDFLuJynwfsR9A_8Troop7DkRG6S8XHzvE&e=>:

> @@ -56,7 +57,7 @@ module mo_rte_lw
   function rte_lw(optical_props, top_at_1, &
                   sources, sfc_emis,       &
                   fluxes,                  &
-                  inc_flux, n_gauss_angles) result(error_msg)
+                  inc_flux, n_gauss_angles, type2stream) result(error_msg)


Since we have only two choices here - two-stream or re-scaled - I would much prefer a logical variable here, like use_2str or similar. (No error checking required!) If the variable is not supplied we assume we will use the re-scaling solution.

________________________________

In rte/mo_rte_lw.F90<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_RobertPincus_rte-2Drrtmgp_pull_60-23discussion-5Fr363490252&d=DwMCaQ&c=birp9sjcGzT9DCP3EIAtLA&r=PGXxSfQJAcousBe_GlGvaQ&m=gyhm3bPHjNJLTWfp9h5_SobTh2dkA1FVII3m-nxUauI&s=2_3cnypCpRdvcg_eB6yQtxJC2-VZpl-RWNW3bNghMQE&e=>:

> @@ -225,6 +250,9 @@ function rte_lw(optical_props, top_at_1, &
                                sfc_emis_gpt, sources%sfc_source,       &
                                gpt_flux_up, gpt_flux_dn)
         !$acc exit data delete(optical_props%tau, optical_props%ssa, optical_props%g)
+        else
+           write(error_msg, "('lw_solver(...ty_optical_props_2str...) processing type canbe (0,1). entered: ', I2)") type2stream


This error checking isn't needed if we use a logical variables - but also please note that we never call write() within the core of rte but instead return strings.

-
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_RobertPincus_rte-2Drrtmgp_pull_60-3Femail-5Fsource-3Dnotifications-26email-5Ftoken-3DAMGWJWUZGLU4R65TCCPOATDQ4OOWRA5CNFSM4J4NG772YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCQZUANA-23pullrequestreview-2D338903092&d=DwMCaQ&c=birp9sjcGzT9DCP3EIAtLA&r=PGXxSfQJAcousBe_GlGvaQ&m=gyhm3bPHjNJLTWfp9h5_SobTh2dkA1FVII3m-nxUauI&s=hic-P4h2yZARA4aiRM8AkiWNXg3ctxf15wM2wpjQKMs&e=>, or unsubscribe<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_notifications_unsubscribe-2Dauth_AMGWJWTVMZHYO7KRBW5JVMDQ4OOWRANCNFSM4J4NG77Q&d=DwMCaQ&c=birp9sjcGzT9DCP3EIAtLA&r=PGXxSfQJAcousBe_GlGvaQ&m=gyhm3bPHjNJLTWfp9h5_SobTh2dkA1FVII3m-nxUauI&s=aYHtKc0_blBRUGTkfj4rO3P7jQNEZx5eYXU0cFzt8zI&e=>.

________________________________

This email is intended solely for the recipient. It may contain privileged, proprietary or confidential information or material. If you are not the intended recipient, please delete this email and any attachments and notify the sender of the error.

RobertPincus:
Please pull my revisions - I changed the integer to a scalar myself. 

I don’t necessarily mind using write() to construct an error message though we haven’t had to do this so far.

inpolonsky:
Robert:

I am curious how you would add no-character variable to the error message in FORTRAN except using write.

I am aware of manipulation based on ASCII table character location but it is a hack - not a standard.

My idea was to provide a user with a meaningful error message.

Regards

Igor

Igor Polonsky
Atmospheric and Environmental Research, Inc.
131 Hartwell Avenue
Lexington, MA  02421-3126
office: 781-761-2228
fax: 781-761-2499
e-mail: ipolonsky@aer.com<mailto:ipolonsky@aer.com>



On 1/8/20 12:16 PM, Robert Pincus wrote:
Please pull my revisions - I changed the integer to a scalar myself.

I don’t necessarily mind using write() to construct an error message though we haven’t had to do this so far.

RobertPincus:
Our error messages are simple and not always informative. When a value is out of bounds, for example, we report this fact rather than the value itself. 

In this case, by making the choice logical, we remove the possibility of users providing invalid values, so we don’t need the error message at all.

RobertPincus:
@inpolonsky For the inadvertent change to `rrtmgp/mo_gas_optics_rrtmgp.F90` perhaps the simplest thing is to check out the version from `develop` and commit it to this branch, i.e. 
```
git checkout tang3
git checkout develop -- rrtmgp/mo_gas_optics_rrtmgp.F90
```
I'm going to suggest some aesthetic changes to the kernels in the interests of consistency and clarity. These should be applied to both `kernels` and `kernels-openacc`

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,295 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+    do igpt=1,ngpt
+      do ilay=1,nlay
+        do icol=1,ncol
+          ssal = ssa(icol, ilay, igpt)
+          wb = ssal*(1._wp - g(icol, ilay, igpt)) / 2._wp
+          scaleTau = (1._wp - ssal + wb )
+          ! Eq.15 of the paper
+          tauLoc(icol, ilay, igpt) = scaleTau * tau(icol, ilay, igpt)
+          ! 
+          ! here ssa is used to store parameter wb/[1-w(1-b)] of Eq.21 of the Tang's paper
```
RobertPincus: Please bring this comment up to date - there's no `ssa` variable used any more

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,295 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+                          top_at_1, Ds_ncol, tauLoc, scaling, &
+                          lay_source, lev_source_inc, lev_source_dec, sfc_emis, sfc_src, &
+                          radn_up, radn_dn)
+
+    flux_up(:,:,:) = flux_up(:,:,:) + weight*radn_up(:,:,:)
+    flux_dn(:,:,:) = flux_dn(:,:,:) + weight*radn_dn(:,:,:)
+  end do
+  end subroutine lw_solver_1rescl_GaussQuad
+
+  pure subroutine scaling_1rescl(ncol, nlay, ngpt, tauLoc, scaling, tau, ssa, g)
```
RobertPincus: Please try to follow the style of the existing code - procedures are set off with long comment lines at the beginning and end

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,295 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+    !$acc exit data delete(tau, ssa, g)
+  end subroutine scaling_1rescl
+
+! -------------------------------------------------------------------------------------------------
+!
+! Longwave no-scattering + a single itergation transport
+!
+! 
+! Implemented based on the paper
+! Tang G, P Yang, GW Kattawar, X Huang, EJ Mlawer, BA Baum, MD King, 2018: Improvement of
```
RobertPincus: In most other places we've simply provide XX et al. and the doi. Of course this doesn't affect functionality; it's about making the code read uniformly.

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,295 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+  real(wp) :: adjustmentFactor
+  if(top_at_1) then
+    !
+    ! Top of domain is index 1
+    !
+    ! 1st Upward propagation
+    do ilev = nlay, 1, -1
+      radn_up(:,ilev) = trans(:,ilev  )*radn_up(:,ilev+1) + source_up(:,ilev)
+      do icol=1,ncol
+         if ( Cn(icol,ilev) > 1e-6 )  then
```
RobertPincus: I've asked about these checks several times before. I'm not convinced they will save time - in a vectorized code the compiler will execute both branches and then mask out the one that isn't needed. I would greatly prefer that we remove all the `if` statements currently in place to save time. You are welcome to leave a comment in place in case we find a big performance it.

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,295 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+                     trans(icol,ilev)*source_dn(icol,ilev) - source_up(icol,ilev) )
+            radn_up(icol,ilev) = radn_up(icol,ilev) + adjustmentFactor
+          endif  
+        enddo  
+    end do
+    ! 2nd Downward propagation
+    do ilev = 1, nlay
+      radn_dn(:,ilev+1) = trans(:,ilev)*radn_dn(:,ilev) + source_dn(:,ilev)
+      do icol=1,ncol
+        if ( Cn(icol,ilev) > 1e-6 )  then
```
RobertPincus: `if` statement

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,295 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+! Implemented based on the paper
+! Tang G, P Yang, GW Kattawar, X Huang, EJ Mlawer, BA Baum, MD King, 2018: Improvement of 
+! the Simulation of Cloud Longwave Scattering in Broadband Radiative Transfer Models, 
+! Journal of the Atmospheric Sciences 75 (7), 2217-2233
+! https://doi.org/10.1175/JAS-D-18-0014.1
+! adustment terms is computed based on solution of the Tang equations
+! for "linear-in-tau" internal source
+! 
+! -------------------------------------------------------------------------------------------------
+subroutine lw_transport_1rescl(ncol, nlay, top_at_1, &
```
RobertPincus: By requiring `An` and `Cn` as inputs you're expecting that users will have called `scaling_1rescl()`. This works in the current implementation, since all roads to `lw_solver_1rescl()` lead through `lw_solver_1rescl_GaussQuad()`, where you are computing the scaling. But when we introduce Jen's optimal angle this might not be true. I don't have a good answer but I wanted to flag the issue

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,295 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+  real(wp), dimension(ncol,nlay+1,ngpt) :: radn_dn, radn_up ! Fluxes per quad angle
+  real(wp), dimension(ncol,       ngpt) :: Ds_ncol
+
+  real(wp), dimension(ncol,nlay,  ngpt) :: tauLoc           ! rescaled Tau
+  real(wp), dimension(ncol,nlay,  ngpt) :: scaling          ! scaling
 
+  integer :: imu, top_level
+  real    :: weight
+  ! Tang rescaling
+  call scaling_1rescl(ncol, nlay, ngpt, tauLoc, scaling, tau, ssa, g)
```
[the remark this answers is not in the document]
inpolonsky: Agree

rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -947,5 +949,295 @@ pure subroutine apply_BC_0(ncol, nlay, ngpt, top_at_1, flux_dn) bind (C, name="a
...
+                             radn_up(:,:,igpt), radn_dn(:,:,igpt))
+
+    call lw_transport_1rescl(ncol, nlay, top_at_1, trans, &
+                             sfc_albedo, source_dn, source_up, source_sfc, &
+                             radn_up(:,:,igpt), radn_dn(:,:,igpt), An, Cn)
+
+  end do  ! g point loop
+end subroutine lw_solver_1rescl
+! -------------------------------------------------------------------------------------------------
+!
```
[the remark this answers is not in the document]
inpolonsky: I reworked the comments

rte/kernels/mo_rte_solver_kernels.F90
inpolonsky: Done

rte/kernels/mo_rte_solver_kernels.F90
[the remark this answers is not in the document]
inpolonsky: I reworked comment

rte/kernels/mo_rte_solver_kernels.F90
inpolonsky: fixed

rte/kernels/mo_rte_solver_kernels.F90
inpolonsky: fixed

rte/kernels/mo_rte_solver_kernels.F90
inpolonsky: replaced with Tang G, et al, 2018: https://doi.org/10.1175/JAS-D-18-0014.1

rte/kernels/mo_rte_solver_kernels.F90
inpolonsky: That particular loop may be non-serializable since the result of the current iteration depends on the previous one. I removed the "if's"

inpolonsky: In scaling_1rescl there also an "if" statement for improbable situation when g=w=1 which cannot ne in LW region. I can remove it with comment. What do you think?

rte/kernels/mo_rte_solver_kernels.F90
inpolonsky: I added comment to lw_solver_1rescl saying that the user must provide the correct scaling to the call of the function

rte/kernels/mo_rte_solver_kernels.F90
RobertPincus: I would prefer to have the input validated before passing it in - so if there will be an error if g = w = 1 when doing the rescaling, let's check for that, probably before even calling the solvers.

inpolonsky: That is a good idea, but it would only identify the potential issue not to solve it.
Currently, for validation I rely on validate_2stream that checks on 
that  0 <=ssa<=1 and -1<=g<=1 which isn't informative in this case.
Should an additional validation function added to the code?

inpolonsky: And actually it not an error it is just an indeterminate form 
w(1-g)/(2-w(1+g))->0/0 ->1 
if w->1 and g->1

inpolonsky: In DISORT or like methods similar problem is solved by allowing 0< =ssa <= 1-epsilon
where epsilon is about 1e-4 or less.

RobertPincus: I'm not sure what the right solution is here. I don't like the `if` statements, but neither do I want code that could crash with a division by zero, even if that only happens when the user has provided bogus input. We could have `validate()` ensure that `ssa < 1._wp-epsilon(1.wp))` but I'm not sure that's a general solution (and anyway I'd want to do the analysis to see what the rounding error is.)

inpolonsky: Actually we have to check only on a single condition that ssa*g < 1. If g=1 - epsilon it means that instead of Tang approximation a correctly rescaled no scattering solution would be more accurate. Let me think how it can be implemented with less pain. Your thoughts would be appreciated. Trivial solution. 
1. create two versions  of the scaling_1rescl
   a)  fast but dangerous
   b)  slow but accurate
2. Check on condition 1-epsilon < ssa*g <=1
   a) if True then call slow
   b) if False call fast

Final thought. I have to verify what solution is accurate if g->1. I suspect that Tang may be no a leader

rte/kernels/mo_rte_solver_kernels.F90
inpolonsky: Actually, the code has 2 types of style:

1.   in  rte/kernels-openacc/mo_rte_solver_kernels.F90  procedures are set off with long comment lines at the beginning and end
2.    in rte/kernels/mo_rte_solver_kernels.F90 procedures are set off with ! of comment blocks

According to your comment your preference is 1 but many files use option 2. 
So I made my code alignment according to existing file style.

PR #66, opened 2020-02-25
Branch develop for clean add-alternate-angle-2 effort

jdelamere:
Copying new code specifically from branch add-alternate-angle into clean branch add-alternate-angle-2.  Note, these copies include only code relevant to the optimal angle effort.

More commits likely to follow, either adding missed changes or repairing code failures.


rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -818,6 +822,8 @@ function load_int(this, available_gases, gas_names, key_species,  &
     real(wp),           dimension(:,:,:,:), intent(in   ) :: planck_frac
     real(wp),           dimension(:,:,:),   intent(in   ), &
                                               allocatable :: rayl_lower, rayl_upper
+    real(wp),           dimension(:,:),     intent(in   ), &
+                                              optional    :: optimal_angle_fit
```
RobertPincus: Is there a reason to make this optional? I don't think you've protected against the case where it isn't supplied

rte/mo_rte_lw.F90
```diff
@@ -178,6 +182,17 @@ function rte_lw(optical_props, top_at_1, &
       return
     end if
 
+    !
+    ! Check extents and values of transport angle secants if supplied
+    !
+    if (present(lw_Ds)) then
+      if(.not. extents_are(lw_Ds, ncol, ngpt)) &
+        error_msg = "rte_lw: lw_Ds inconsistently sized"
+      if(.not. any_vals_less_than(lw_Ds, 0._wp)) &
```
RobertPincus: I think you don't want this not: you're checking that all the Ds are > 0, no?

rte/mo_rte_lw.F90
```diff
@@ -214,12 +229,23 @@ function rte_lw(optical_props, top_at_1, &
         !$acc enter data copyin(optical_props%tau)
         error_msg =  optical_props%validate()
         if(len_trim(error_msg) > 0) return
-        call lw_solver_noscat_GaussQuad(ncol, nlay, ngpt, logical(top_at_1, wl), &
-                              n_quad_angs, gauss_Ds(1:n_quad_angs,n_quad_angs), gauss_wts(1:n_quad_angs,n_quad_angs), &
+
+        if (present(lw_Ds)) then
+          call lw_solver_noscat(ncol, nlay, ngpt, &
+                                logical(top_at_1, wl), lw_Ds, gauss_wts(1,1), &
```
RobertPincus: As an aesthetic choice, can the layout of the lines in the two branches here be the same, i.e. logical(top_at_1, wl) sits on its own line in both branches?

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -1382,6 +1401,64 @@ function get_col_dry(vmr_h2o, plev, latitude) result(col_dry)
...
+    character(len=128)                           :: err_msg
+    !----------------------------
+    integer  :: ncol, nlay, ngpt, nbnd
+    integer  :: icol, ilay, igpt, bnd
+    real(wp) :: t, trans_total(optical_props%get_ncol(), optical_props%get_ngpt())
+    !----------------------------
+    ncol = optical_props%get_ncol()
+    nlay = optical_props%get_nlay()
+    ngpt = optical_props%get_ngpt()
+    nbnd = optical_props%get_nband()
```
RobertPincus: Is variable nbnd used?

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -127,6 +127,7 @@ module mo_gas_optics_rrtmgp
                                                                ! planck_frac(g-point, eta, pressure, temperature)
     real(wp), dimension(:,:),     allocatable :: totplnk       ! integrated Planck irradiance by band; (Planck temperatures,band)
     real(wp)                                  :: totplnk_delta ! temperature steps in totplnk
+    real(wp), dimension(:,:),     allocatable :: optimal_angle_fit
```
RobertPincus: Can you add the dimensions to the comments as per other arrays?

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -1382,6 +1401,64 @@ function get_col_dry(vmr_h2o, plev, latitude) result(col_dry)
...
+  ! Compute a transport angle that minimizes flux errors at surface and TOA based on empirical fits
+  !
+  function compute_optimal_angles(this, optical_props, optimal_angles) result(err_msg)
+    ! input
+    class(ty_gas_optics_rrtmgp),  intent(in   ) :: this
+    class(ty_optical_props_arry), intent(in   ) :: optical_props
+    real(wp), dimension(:,:),     intent(  out)  :: optimal_angles
+    character(len=128)                           :: err_msg
+    !----------------------------
+    integer  :: ncol, nlay, ngpt, nbnd
```
RobertPincus: nbnd isn't used?

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -1382,6 +1401,64 @@ function get_col_dry(vmr_h2o, plev, latitude) result(col_dry)
...
+        !
+        t = 0._wp
+        do ilay = 1, nlay
+          t = t + optical_props%tau(icol,ilay,igpt)
+        end do
+        trans_total(icol,igpt) = exp(-t)
+      end do
+    end do
+
+    !$acc parallel loop gang vector collapse(2)
```
RobertPincus: Is there a reason to have these two loops separate? Seems like it'd be both easy and clear (and memory friendly) to combine the calculation of total column transmission with the calculation of optimal angle.

rte/mo_rte_lw.F90
```diff
@@ -178,6 +182,17 @@ function rte_lw(optical_props, top_at_1, &
       return
     end if
 
+    !
+    ! Check extents and values of transport angle secants if supplied
+    !
+    if (present(lw_Ds)) then
```
RobertPincus: You will want to ensure that `lw_Ds` are supplied only if `optical_props` is of type `_1scl`.

rte/mo_rte_lw.F90
jdelamere: I did not add this code...says "blame"...From Eli, "we are doing a linear function of T.  so the values for T=0 and T=1 have to be reasonable.  T=0 is the second coefficient and T=1 is their sum.  so one could check that the second and the sum are both ge > 1.  i can think of no circumstances in which they’d be > 2, so I suppose one could also check that both the first and the sum are < 2."

jdelamere: Eli also says, "i’m assuming that these are just sanity checks for bad values, so my proposed tests, although soft, will find insane values.  if we wanted tighter tests, then it looks like having both the second and the sum between 1.5 and 2 would work."

rrtmgp/mo_gas_optics_rrtmgp.F90
jdelamere: Not used.  Removing.

rte/mo_rte_lw.F90
RobertPincus: My suggestion is not to be too clever, but simply check that all `lw_Ds` are > 0.

jdelamere: I think the point is some of the lw_Ds will be 0 or less than 0.  I'll add eli's logic.

RobertPincus: @jdelamere The lw_Ds are the secants of the propagation angle, one per g-point and column. Eli's logic refers to coefficient of linear fit that produce these secants and doesn't apply here. Or have I missed something?

rte/mo_rte_lw.F90
jdelamere: Robert - can you help me with the logic we should have in this part of the code? We now have use_2stream and lw_Ds as arguments to rte_lw.  But they aren't used at the same time....

jdelamere: okay - I will add these checks.

RobertPincus: It should be an error if `use_2stream` is supplied but the type of optical_props is not `_2str` is of course what I meant

rte/mo_rte_lw.F90
jdelamere: Minor question here regarding style - you have routines "any_vals_less_than" - if we need less than, or equal to, do we set up a new set of functions?

RobertPincus: If you're testing the secant it seems like "less than 1" works, but you could also check for any_vals_less_than( val + spacing(val)). That's probably preferable to developing a whole new set of routines.

RobertPincus:
@jdelamere Just looking over your shoulder at the current changes... please don't use logic that encodes the type of the optical properties as a 4-character string you test later. That's indirect and will confuse folks. There's already a section of `rte_lw()`  in decisions are made based on this type. It's aesthetically pleasing to put all the checks together but not crucial. You can also complain about the presence of `lw_Ds` if it turns out that people are trying to use them in scattering calculations.

rte/mo_rte_lw.F90
jdelamere: I knew that code I had in there was no good but I do want to test before hand - what if we add more scenarios to rte_lw? For example, if I'm in the 1scl code block -> then I have to check that "use_2stream" is false and any other optional variables that we might add later on rather than check that is use_2stream is positive, the class must be 2str.  Well, I'll change it because that coding was bad but illustrated the point - hence my request for you to check it.

PR #67, opened 2020-02-25
Sj.v1

inpolonsky:
please merge the latest surface Jacobian


RobertPincus:
@inpolonsky Note that the GPU runs are still failing in the continuous integration. I'd like to wait to review the code until it passes all the tests.

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -561,8 +561,12 @@ subroutine compute_Planck_source(                        &
     real(wp), dimension(ngpt,     ncol), intent(out) :: sfc_src
     real(wp), dimension(ngpt,nlay,ncol), intent(out) :: lay_src
     real(wp), dimension(ngpt,nlay,ncol), intent(out) :: lev_src_inc, lev_src_dec
+
+    real(wp), dimension(ngpt,     ncol), intent(out) :: sfc_source_Jac
     ! -----------------
     ! local
+    real(wp), parameter                             :: dST = 1.0_wp
```
RobertPincus: Can this variable name be made more explicit, e.g. delta_Tsurf?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -596,15 +601,23 @@ subroutine compute_Planck_source(                        &
     !
     !$acc parallel loop
     do icol = 1, ncol
-      call interpolate1D(tsfc(icol), temp_ref_min, totplnk_delta, totplnk, planck_function(1:nbnd,1,icol))
+      call interpolate1D(tsfc(icol)      , temp_ref_min, totplnk_delta, totplnk, planck_function(1:nbnd,1,icol))
+      call interpolate1D(tsfc(icol) + dST, temp_ref_min, totplnk_delta, totplnk, planck_function(1:nbnd,2,icol))
     end do
+    !$acc parallel loop collapse(2)
```
RobertPincus: Would it make sense to fuse this loop with the one below and write sfc_source_Jac(igpt,icol) = pfrac(igpt,sfc_lay,icol) * (planck_function(gpoint_bands(igpt), 2, icol) - planck_function(gpoint_bands(igpt), 1, icol)) 
?

rrtmgp/kernels/mo_gas_optics_kernels.F90
```diff
@@ -504,8 +504,12 @@ subroutine compute_Planck_source(                        &
     real(wp), dimension(ngpt,     ncol), intent(out) :: sfc_src
     real(wp), dimension(ngpt,nlay,ncol), intent(out) :: lay_src
     real(wp), dimension(ngpt,nlay,ncol), intent(out) :: lev_src_inc, lev_src_dec
+
```
RobertPincus: All comments from OpenACC version of kernels apply here too

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -501,6 +503,7 @@ function compute_gas_taus(this,                       &
     ! Compute dry air column amounts (number of molecule per cm^2) if user hasn't provided them
     !
     idx_h2o = string_loc_in_array('h2o', this%gas_names)
+    col_dry_wk => col_dry_arr
```
RobertPincus: Not that it matters, since pointer assignment is cheap, but I wonder why you do this since you might re-assign the pointer just below.

rte/kernels-openacc/mo_rte_solver_kernels.F90
```diff
@@ -1238,6 +1266,7 @@ subroutine lw_solver_1rescl(ncol, nlay, ngpt, top_at_1, D,
                                              trans       ! transmissivity  = exp(-tau)
     real(wp), dimension(ncol,nlay,ngpt) :: source_dn, source_up
     real(wp), dimension(ncol,     ngpt) :: source_sfc, sfc_albedo
+    real(wp), dimension(ncol,nlay+1,ngpt) :: dummy
```
RobertPincus: What is there a dummy variable? Seems like we could delete it with no problem

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -774,6 +784,20 @@ function source(this,                               &
     call reorder123x321(lay_source_t, sources%lay_source)
     call reorder123x321(lev_source_inc_t, sources%lev_source_inc)
     call reorder123x321(lev_source_dec_t, sources%lev_source_dec)
+    if (present(sourcesJac)) then
+      !$acc enter data copyin(sourcesJac)
+      !$acc enter data create(sourcesJac%sfc_source)
+      !$acc parallel loop collapse(2)
```
RobertPincus: Before doing this loop we should check that sourcesJac has the same spectral discretization and number of column as the optical properties

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
inpolonsky: Sure. Not a problem. Done

RobertPincus: If you have done this it's not yet visible - did you push your changes?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
inpolonsky: Agreed. Done

rrtmgp/kernels/mo_gas_optics_kernels.F90
inpolonsky: Done.

rrtmgp/mo_gas_optics_rrtmgp.F90
inpolonsky: Because, in the original code result in a crush when was run on GPU with error message "Null pointer for col_dry_wk". I was trying to nullify it but with no success. That was the simplest solution and I am still puzzled by the problem.

rte/kernels-openacc/mo_rte_solver_kernels.F90
inpolonsky: That is a problem I am facing. Since lw_transport_noscat requires non optional radn_upJac. I created a dummy variable. For the sake of clarity, I renamed it to dummyRadn_upJac

rrtmgp/mo_gas_optics_rrtmgp.F90
inpolonsky: I placed the check into function gas_optics_int where  check on sources was performed

PR #77, opened 2020-05-27
A few small cosmetic fixes

Chiil:
1. Make the `Makefile.conf` for gfortran use -DUSE_CBOOL by default in order make it compile with the `std=f2003` flag.
2. Corrected a typo in the output to the screen in the RFMIP test case and completed the list of required packages in the Readme.md of the example

### DIFF
--- a/build/Makefile.conf.gfortran
+++ b/build/Makefile.conf.gfortran
@@ -9,9 +9,9 @@ export FC = gfortran
 #
 # Optimized
 #
-export FCFLAGS = -ffree-line-length-none -m64 -std=f2003 -march=native -O3
+export FCFLAGS = -ffree-line-length-none -m64 -std=f2003 -march=native -O3 -DUSE_CBOOL
 #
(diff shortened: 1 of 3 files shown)


skosukhin:
Just a notice. We don't need `toolz` itself. What we need is `dask.array` but it doesn't work without `toolz`. It is possible that the list of dependencies for `dask.array` will change in the future.

Chiil:
For me, a `pip install dask` was insufficient as it did not install the `toolz` dependency.

skosukhin:
@Chiil please, take a look [here](https://docs.dask.org/en/latest/install.html#pip). As I said, the list of dependencies for particular modules might change in the future. We need `dask.array`, therefore the users are recommended to run `pip install "dask[complete]"` or `pip install "dask[array]"`.

Chiil:
Alright, `pip install dask[array]` works indeed, thanks. It could be good to include that message for the naive `pip` users (including me).

PR #79, opened 2020-06-09
Loop order change and g = 0. assignment outside of loop. ~5x speedup in

Chiil:
Change in loop order and removal of `g` assignment in loop makes performance approximately five times faster for this kernel.


RobertPincus:
Closed by 972073aaae7ac47581f6140f8b8e52877de2a0d6

PR #82, opened 2020-06-24
fix to avoid overflow in minor gas code when using single precision, …

peterukk:
…plus faster find_gas using the findloc intrinsic


peterukk:
Tests are failing with gfortran-8, because apparently this Fortran 2008 feature is only supported since gfortran 9.0

RobertPincus:
@peterukk The CI tests are also failing with PGI compilers on Piz Daint. Perhaps it's better to remove the `findloc()` optimization and just retain the precision-preserving refactoring?

peterukk:
Sure. The find_gas function accounted for 2.5% of RRTMGP's runtime with a block size of 18, and 4.9% with a block size of 10, so it's not super important. For just implementing the single precision fix, is that something you can from this pull request or should I make another one?

peterukk:
Oops, there was a typo in the gas optics code. I'll just make a new pull request, its cleaner.

PR #89, opened 2020-09-11
Working CI onto develop from feature-enhanced-ci

RobertPincus:


RobertPincus:
Closed via 201e5f4, although ifort packages have changed and netCDF library isn't compiling because the C compiler is broken.

PR #99, opened 2021-01-06
Support GPUs with OpenMP 4.5 

naromero77:
Tested and passing CI with the Cray Compiler 10.x compiler and NVidia GPUs.


examples/all-sky/rrtmgp_allsky.F90
```diff
@@ -1,390 +1,422 @@
...
-    if(write_fluxes) call write_lw_fluxes(input_file, flux_up, flux_dn)
-    !$acc exit data delete(t_sfc, emis_sfc)
-  else
-    !$acc exit data copyout(flux_up, flux_dn, flux_dir)
-    if(write_fluxes) call write_sw_fluxes(input_file, flux_up, flux_dn, flux_dir)
-    !$acc exit data delete(sfc_alb_dir, sfc_alb_dif, mu0)
-  end if
-  !$acc enter data create(lwp, iwp, rel, rei)
-end program rte_rrtmgp_clouds
+subroutine stop_on_err(error_msg)
```
RobertPincus, line 1: In every other file the diffs show only the addition of the OpenMP directives. This files shows up as completely changed, which is confusing. Do you know why?

examples/rfmip-clear-sky/Makefile
```diff
@@ -53,7 +53,7 @@ rrtmgp_rfmip_lw:   rrtmgp_rfmip_lw.o   $(ADDITIONS) $(RRTMGP_BUILD)/librte.a $(R
 
 rrtmgp_rfmip_lw.o: rrtmgp_rfmip_lw.F90 $(ADDITIONS)
 
-rrtmgp_rfmip_sw:   rrtmgp_rfmip_sw.o   $(ADDITIONS) $(RRTMGP_BUILD)/librte.a $(RRTMGP_BUILD)/librrtmgp.a
+rrtmgp_rfmip_sw:   rrtmgp_rfmip_sw.o   $(ADDITIONS) $(RRTMGP_BUILD)/librrtmgp.a $(RRTMGP_BUILD)/librte.a
```
RobertPincus, line 56: I realize this is harmless but why are the libraries reordered here?

examples/rfmip-clear-sky/Makefile
```diff
@@ -5,8 +5,8 @@ RRTMGP_BUILD = $(RRTMGP_ROOT)/build
 #
 # RRTMGP library, module files
 #
-LDFLAGS   += -L$(RRTMGP_BUILD)
-LIBS      += -lrrtmgp -lrte
+# LDFLAGS   += -L$(RRTMGP_BUILD)
```
RobertPincus, line 8: Why are these removed?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -214,27 +222,41 @@ subroutine compute_tau_absorption(                &
...
     ! ---------------------
     ! Layer limits of upper, lower atmospheres
     ! ---------------------
     top_at_1 = play(1,1) < play(1, nlay)
     if(top_at_1) then
       !$acc parallel loop
+      !$omp target teams distribute parallel do simd
       do icol = 1,ncol
         itropo_lower(icol,2) = nlay
+#ifdef _CRAYFTN
```
RobertPincus, line 238: Not pretty but effective. Do you know if Cray plans to support the `minloc` and `maxloc` intrinsics on the device at some point?

rrtmgp/mo_gas_concentrations.F90
```diff
@@ -142,18 +143,27 @@ function set_vmr_scalar(this, gas, w) result(error_msg)
...
     if (.not. associated(this%concs(igas)%conc)) then
       allocate(this%concs(igas)%conc(1,1))
       !$acc enter data create(this%concs(igas)%conc)
+      !$omp target enter data map(alloc:this%concs(igas)%conc)
     end if
 
+    p => this%concs(igas)%conc(:,:)
     !$acc kernels
+    !$omp target map(to:w)
+#ifdef _CRAYFTN
```
RobertPincus, line 160: Is the coding especially tricky here, or is this a weakness/bug in the Cray compiler?

rte/mo_rte_util_array.F90
```diff
@@ -75,9 +79,28 @@ logical function any_vals_less_than_3D(array, check_value)
 
     real(wp) :: minValue
 
+#ifdef _OPENMP
```
RobertPincus, line 82: Here, too - will Cray eventually support `minval` and `maxval` intrinsics?

tests/Makefile
```diff
@@ -5,8 +5,8 @@ RRTMGP_BUILD = $(RRTMGP_ROOT)/build
 #
 # RRTMGP library, module files
 #
-LDFLAGS   += -L$(RRTMGP_BUILD)
-LIBS      += -lrrtmgp -lrte
+# LDFLAGS   += -L$(RRTMGP_BUILD)
```
RobertPincus, line 8: Here too I'm confused as to why the Makefiles are being modified

RobertPincus (commented):
Thanks so much for this. The next step, I guess, is to get CI working in this repo.

examples/all-sky/rrtmgp_allsky.F90
naromero77: Correct, its because of the carriage returns. I have a specific commit for this already in this PR:
https://github.com/earth-system-radiation/rte-rrtmgp/pull/99/commits/c03b95f4acc24028070c0ab7b5ab16ffd651110e

Note that when those `^M` are removed it will always look like the entire file changed.

Unfortunately, this file has become a headche because of the `^M`. (I spent almost half a day dealing with the best way to handle it already).

I think the cleanest way to do this is as follows. 
1. close this PR
2. create a separate PR for the control `^M`
3. Merge the PR in for the control `^M`
3. rebase/merge my changes into the tip of the new develop branch
4. create a new PR

Another possibility, would be:
1. close this PR
2. Rebase my feature branch to remove the commit with the control `^M`.
2. This will cause issues with other commits being applied, but hopefully this won't be too bad.
3. Push the changes into my feature branch

Let me know what you would like to do.

examples/rfmip-clear-sky/Makefile
naromero77: This was the recommended fix from the Cray scientist that I worked with.

As far as I can tell, this is a feature/weakness of the Cray linker and other linkers that are partly-based on LLVM and support OpenMP for GPUs. They are just not as robust as their CPU brethren. If the static linkage is not in the right order, the binary cannot be built.

examples/all-sky/rrtmgp_allsky.F90
RobertPincus: You have figured out how to remove the `^M` characters now? If you explain how I could update my branch, so that the PR shows only the OpenMP additions.

examples/rfmip-clear-sky/Makefile
naromero77: The Makefile is linking these static libraries twice: librrtmgp and librte.a

Firstly here:
```
rrtmgp_rfmip_sw:   rrtmgp_rfmip_sw.o   $(ADDITIONS) $(RRTMGP_BUILD)/librrtmgp.a $(RRTMGP_BUILD)/librte.a`
```

then secondly here:
```
LDFLAGS   += -L$(RRTMGP_BUILD)
LIBS      += -lrrtmgp -lrte
```

Again, this is a feature/weakness of the Cray linker and other linkers that are partly-based on LLVM and support OpenMP for GPUs. They are just not as robust as their CPU brethren. Because of the duplicate linkage, the binary cannot be built. 

So, the take home message is that the GPU linker is less robust than the CPU linker. The static libraries can only be linked once and must be in a specific order.

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
naromero77: I made the request almost a year ago, but sadly it has not been implemented yet.

I have made this request to other vendors like Intel and they will support this early on.

rrtmgp/mo_gas_concentrations.F90
naromero77: Yes. This is the most problematic file to create an OpenMP GPU version. 

The PGI OpenACC compiler is miles ahead of the Cray OpenMP compiler here. Suppose you have a derived type `foo` which contains another derived type `bar` and several allocatable components. It simply cannot handle mapping `foo%bar%comp` to the GPU. It doesn't really know how to handle that. I think the OpenMP compilers will get better at this in the future though.

rte/mo_rte_util_array.F90
naromero77: `minval` and `maxval` are supported on the GPU. The problem here has to do with performance.

Sadly, OpenMP 4.5 - 5.0, does not get Fortran array notation and there is no equivalent to `$!acc kernel`. So the loops have to be explicit to expose the parallelism. Thus, you have to work with the scalar versions, i.e. `min` and `max`.

I avoided writing something custom, but without these customized routines. The allsky benchmark with tons of columns does not see any speed-up on the GPU.

tests/Makefile
naromero77: Its the duplicate static linking problem that the GPU linker cannot handle. See previous comment.

examples/all-sky/rrtmgp_allsky.F90
RobertPincus: @naromero77 This question is for you

naromero77: On Linux it is super easy with this utility. 
`https://linux.die.net/man/1/dos2unix`

If you are on Mac and use homebrew, do:
`brew install dos2unix`

It should be as easy as: `dos2unix rrtmgp_allsky.F90 rrtmgp_allsky.F90` should do it.

naromero77: Our last messages crossed paths -- let me know what you would like me to do next here.

If you update your develop branch with the `^M` change, it *might* create a conflict on this PR. It is also possible the git will be smart enough to figure out how to do the merge cleanly an only show the OpenMP changes in the PR which is what you want. I have run into this issue in the past before and unfortunately git-merge had a tough time with it.

m214089:
A solution for your line-ending problem is part of git. Take a look at:

https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings

examples/all-sky/rrtmgp_allsky.F90
naromero77: Alternatively, if its just a matter of reviewing the OpenMP changes, then most of the changes for this `rrtmgp_allsky.f90` are in this commit:
https://github.com/earth-system-radiation/rte-rrtmgp/pull/99/commits/0e937fe5081da526ae2e81f6dbc9fd4bd8764c6f

And a few more here:
https://github.com/earth-system-radiation/rte-rrtmgp/pull/99/commits/35e6997b3c32ebe269d0a2e0a9bb522ac5ac6f9f

naromero77:
Thanks for the comment, this is helpful beforehand to prevent issues but I am not sure how this helps after the fact.
> A solution for your line-ending problem is part of git. Take a look at:
> 
> https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings

examples/all-sky/rrtmgp_allsky.F90
RobertPincus: Well, despite having `dos2unix`ed the file in a separate commit on my end the merge is still showing the entire file as changed. Shrug.

naromero77: Another idea that I have.  Do not change anything on your end, but I will try deleting the `^M` specific commit in my feature branch and then do a rebase. I can try that on a local test branch and see if it works first. Worse case scenario is that it could confuse the compiler during the parsing process and lead to a CI failure.

naromero77: But anytime you remove `^M`, it will show the entire file as changed. There is no way around that AFAIK.

naromero77:
This is a note to confirm that all CI is passing on my end, this includes both the GitHub CI and the NERSC CI. Check it out here:
https://github.com/naromero77/rte-rrtmgp/tree/omp-45-cray-sync-0d01651

Perhaps try restarting the GitHub CI on `feature-openmp-gpus` a couple of times to see if it passes here.

RobertPincus (approved):
Thanks, this looks good.

PR #102, opened 2021-03-23
OpenMP GPU fixes for the Cray 11.x compiler

naromero77:
These commits enable the clear_sky_regression test to run correctly with Cray 11.x compiler with the NVidia V100 backend. Untested on the P100.


RobertPincus (approved):
Looks great Nichols, thanks very much for working this through. Delighted to have the OpenMP implementation running within continuous integration on the CCE compiler.

PR #104, opened 2021-03-24
Fix a floating point literal

brhillman:
Use explicit kind with PaTohPa floating point literal to be sure of
precision. This was needed for bit-for-bit consistency between the
fortran kernels and a C++ port.


RobertPincus (approved):
Thanks for the catch, @brhillman

brhillman:
Wow, that was a fast merge, thanks @RobertPincus!

PR #112, opened 2021-05-05
Improvements to make more comformant to OpenMP spec

naromero77:
The Intel oneAPI compiler (which supports the Intel GPU) has more strict compile-time checks for compliance with the latest OpenMP spec. The Cray compiler will actually compile the code that contains map clauses of discontiguous arrays, but will then emit a run-time error stating that it cannot map discontiguous arrays.


RobertPincus (approved):
Sorry you have to deal with this level of fussiness. Thanks for the contribution.

PR #114, opened 2021-05-05
Re-write loop to support NVidia's Fortran compiler.

naromero77:
Current NVidia Fortran compiler does not support private clause on arrays in an OpenMP target region. This can be avoided by using a scalar and re-writing the loop as suggested by @RobertPincus 

We are now able to build with the NVidia Fortran compiler, but still not able to run.


RobertPincus (approved):
Sweet. Curious why it doesn't run yet.

PR #115, opened 2021-05-10
Installation procedure improved

MHBalsmeier:
Hi,

I do not know if this is useful, but when compiling RRTMGP on Linux Ubuntu, I encountered some problems and fixed them. These were:

- the variables RRTMGP_ROOT and NFHOME in the Makefiles were not set properly
- modern systems often do not find python anymore, therefore I replaced the command python by python3
- in some files some lines were too long

I do not know if this is useful, but I think it is unlikely I am the only one who encountered these problems.


RobertPincus:
@MHBalsmeier It's not particularly robust to explicitly name a version of Python (e.g. to use `python3`). If you have Python3 on your system you can make it the default with `apt install python-is-python3` as we do in the [containers we use](
https://github.com/earth-system-radiation/containers/blob/030e4382c865f35eacf0fc6bfb561f957085f039/Dockerfile-add-python#L9) for continuous integration. 

In general if you're having difficulty with a local installation you might look to see how the continuous integration is set up (e.g. in .github/workflows/continuous-integration.yml)

MHBalsmeier:
Okay, I undid this, now only the long lines remain.

RobertPincus:
@MHBalsmeier Great. Finally, if you be so kind as to follow the instructions in `Contributing.md` and target the PR to the `develop` branch instead of `main`  I will be happy to merge the changes in.

RobertPincus:
Mmmm, this PR has conflicts I don't understand and can't be merged. I'll commit separately.

PR #116, opened 2021-05-11
Unused local variables removed, conversion warnings resolved

MHBalsmeier:
Hi, I removed all local variables that were never used and an unused function (write_string in mo_simple_netcdf). Additionally, I conversed some types to int in the subroutine load_and_init (resulted in conversion warnings).


RobertPincus (approved):
Thanks for these simplifications. It's good that you didn't remove every function which is not used as some are there to grow into especially in the optical properties code.

PR #117, opened 2021-05-21
Fixes from ICON

alexeedm:
I've changed OMP pragmas to match OpenACC in a few places, would be good if @naromero77 checks these places.


extensions/mo_rrtmgp_clr_all_sky.F90
```diff
@@ -85,6 +85,8 @@ function rte_lw(k_dist, gas_concs, p_lay, t_lay, p_lev,    &
     ngpt  = k_dist%get_ngpt()
     nband = k_dist%get_nband()
 
+    !$acc update if_present host(p_lay(1,1), p_lay(1,nlay:nlay))
+    !$omp target update from(p_lay(1,1), p_lay(1,nlay:nlay))
```
naromero77, line 89: Why not just the `update` the entire matrix? Is it too large? Principle of least surprise -- that way you don't end up with a partially update matrix on the host.

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -67,14 +67,14 @@ subroutine interpolation( &
     ! local indexes
     integer :: icol, ilay, iflav, igases(2), itropo, itemp
 
-    !$acc enter data copyin(flavor,press_ref_log,temp_ref,vmr_ref,play,tlay,col_gas)
+    !$acc data copyin(flavor,press_ref_log,temp_ref,vmr_ref,play,tlay,col_gas) &
+    !$acc      copyout(jtemp,jpress,tropo,jeta,col_mix,fmajor,fminor) &
+    !$acc      create(ftemp,fpress)
```
naromero77, line 72: Why the change from unstructured to structured data region? It is it trying to get around a compiler bug?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
```diff
@@ -668,7 +669,7 @@ subroutine compute_Planck_source(                        &
     ! Explicitly unroll a time-consuming loop here to increase instruction-level parallelism on a GPU
     ! Helps to achieve higher bandwidth
     !
-    !$acc parallel loop collapse(3)
+    !$acc parallel loop present(planck_function) collapse(3)
```
naromero77, line 672: `planck_function` is a function. What does this clause do since the function should always be present? I thought only variables would be allowed as arguments.

extensions/mo_rrtmgp_clr_all_sky.F90
alexeedm: It is way too large to update the full matrix. Maybe there's a better way to compute top_at_1, or get it as an input to top-level radiation routines, @RobertPincus what do you think?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
alexeedm: Despite the name, `planck_function` is an array :) See line 606

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -646,8 +649,8 @@ function compute_gas_taus(this,                       &
             idx_h2o, col_dry_wk,col_gas, &
             fminor,jeta,tropo,jtemp,     & ! local input
             tau_rayleigh)
-      !$acc exit data detach(col_dry_wk) delete(this%krayl)
-      !$omp target exit data map(from:col_dry_wk) map(release:this%krayl)
+      !$acc exit data delete(this%krayl)
+      !$omp target exit data map(release:this%krayl)
```
naromero77, line 653: OK, good change. Getting rid of the `detach` on `col_dry_wk` should help with OpenMP v. OpenACC compatibility.

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
alexeedm: Actually, at a later point in time I'd like to change to structured data sections whenever possible. It's simpler, less error-prone and potentially more efficient.
In this particular place structured section helps reduce the code a bit. Variables with indirect indexing (like `temp_ref` or `vmr_ref`) should better be explicitly listed as `present` in the kernels. Compiler may not be able to identify which part of the variable it has to apply implicit data sections to, and this may lead to a bug. I latest NVHPC version work fine in this case, but I want to make sure that's the case.
Then I just wanted to combine the said annotation in the kernel with the allocation/deallocation, which becomes exactly the structured data section.

extensions/mo_rrtmgp_clr_all_sky.F90
naromero77: While it gets the job done, my main concern is that this could surprise someone unfamiliar with the code. The GPU has the updated matrix, but the CPU only has two matrix elements updated. I supposed it would be OK if you add a comment/warning in the code explaining why this is done in this particular way.

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
naromero77: I agree that structure OpenMP data regions are general much less error prone. I think it would be good to change the OpenMP offload to minimize divergence. Could it get done in this pull request?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
naromero77: My fault -- I misread the code.

rte/kernels-openacc/mo_rte_solver_kernels.F90
```diff
@@ -198,7 +196,7 @@ subroutine lw_solver_noscat(ncol, nlay, ngpt, top_at_1, D, weight,
     !
     ! Surface reflection and emission
     !
-    !$acc parallel loop collapse(2)
+    !$acc parallel loop collapse(2) no_create(gpt_Jac)
```
naromero77, line 199: Mostly just trying to understand the code. does `no_create(gpt_Jac)` have a performance benefit?

naromero77:
@RobertPincus Is PizDaint offline?

rrtmgp/kernels-openacc/mo_gas_optics_kernels.F90
alexeedm: Sure, I'll update it after the weekend

extensions/mo_rrtmgp_clr_all_sky.F90
alexeedm: I can change to a bit different approach then. I can place the comparison into the `acc kernels` construct and copyout the result. Does OpenMP provide something similar to `acc kernels`, or just an offload parallel region with just 1 thread?

rte/kernels-openacc/mo_rte_solver_kernels.F90
alexeedm: `no_create` wouldn't create the `gpt_Jac` array on the GPU if it has not been allocated. So like `present if (acc_present(gpt_Jac))`. So in case we don't need the Jacobians, we won't allocate the said array on the device in this kernel.

extensions/mo_rrtmgp_clr_all_sky.F90
naromero77: The closet thing to `acc kernels` is `omp declare target`, but this is mostly used for calling small user-defined functions from inside a target region. If you just need one thread to do it on the GPU, you can just do `$omp start target` and `$omp end target`. Here is an example:
https://github.com/earth-system-radiation/rte-rrtmgp/blob/main/rrtmgp/mo_gas_optics_rrtmgp.F90#L738-L743

alexeedm: Great, that should do. Thanks!

rte/kernels-openacc/mo_rte_solver_kernels.F90
naromero77: Basically, `gpt_Jac` is an optional argument somewhere higher up in the calls stack and so it may not exist at all on the CPU.

OK, I am less sure how to do this in OpenMP offload or even what the OpenMP runtime will do. Normally, implicit mapping rules would say to map the array even if not explictly mapped. But here you have a call path that is only executed, if `gpt_Jac` exists. https://github.com/earth-system-radiation/rte-rrtmgp/blob/026cd9d81ce228f2f66db28c0f47c0cec89a7ab4/rte/kernels-openacc/mo_rte_solver_kernels.F90#L208-L210

I am not sure how to do this with OpenMP offload or maybe it is fine that way it is. Let me ask around.

naromero77: I don't think the OpenMP offload version of this code will work. I remember talking to @RobertPincus about this because we found some odd bugs in the Cray compiler. We drafted something, but never committed the changes.

I think we need to do something like this
```
 if(do_Jacobians)
   !$acc parallel loop collapse(2)
    !$omp target teams distribute parallel do simd collapse(2)
    do igpt = 1, ngpt
      do icol = 1, ncol
        !
        ! Surface albedo, surface source function
        !
        radn_up   (icol,sfc_level,igpt) = radn_dn(icol,sfc_level,igpt)*(1._wp - sfc_emis(icol,igpt)) + &
                                          sfc_src(icol,          igpt)*         sfc_emis(icol,igpt)
        gpt_Jac(icol,sfc_level,igpt) = sfc_srcJac(icol,        igpt)*         sfc_emis(icol,igpt)
      end do
    end do
else
   !$acc parallel loop collapse(2)
    !$omp target teams distribute parallel do simd collapse(2)
    do igpt = 1, ngpt
      do icol = 1, ncol
        !
        ! Surface albedo, surface source function
        !
        radn_up   (icol,sfc_level,igpt) = radn_dn(icol,sfc_level,igpt)*(1._wp - sfc_emis(icol,igpt)) + &
                                          sfc_src(icol,          igpt)*         sfc_emis(icol,igpt)
      end do
    end do
end
```
but its ugly to have to have duplicated code.

RobertPincus: @naromero77 Seems like we should just have separate kernels: 
```suggestion
   !$acc parallel loop collapse(2)
    !$omp target teams distribute parallel do simd collapse(2)
    do igpt = 1, ngpt
      do icol = 1, ncol
        !
        ! Surface albedo, surface source function
        !
        radn_up   (icol,sfc_level,igpt) = radn_dn(icol,sfc_level,igpt)*(1._wp - sfc_emis(icol,igpt)) + &
                                          sfc_src(icol,          igpt)*         sfc_emis(icol,igpt)
      end do
    end do

 if(do_Jacobians)
   !$acc parallel loop collapse(2)
    !$omp target teams distribute parallel do simd collapse(2)
    do igpt = 1, ngpt
      do icol = 1, ncol
        !
        ! Surface albedo, surface source function
        !
        gpt_Jac(icol,sfc_level,igpt) = sfc_srcJac(icol,        igpt)*         sfc_emis(icol,igpt)
      end do
    end do
end if 
```

naromero77: yes, that would be safer IMO.

extensions/mo_rrtmgp_clr_all_sky.F90
RobertPincus: @alexeedm Will you update this?

RobertPincus (approved):
Would be great to resolve @naromero77's outstanding suggestions for the OpenMP sections but I will merge this so we can move on

RobertPincus:
@alexeedm It seems the OpenMP is now broken: https://github.com/earth-system-radiation/rte-rrtmgp/runs/2714670788?check_suite_focus=true I would prefer this be working before merging into `develop`

alexeedm:
Looking into this right now.

extensions/mo_rrtmgp_clr_all_sky.F90
alexeedm: resolved

alexeedm:
I'll try to understand the OMP issue on Piz Daint. NVHPC can't compile the OMP offload code that we have

rte/mo_rte_lw.F90
```diff
@@ -220,7 +220,7 @@ function rte_lw(optical_props, top_at_1, &
     !
     ! Ensure values of tau, ssa, and g are reasonable if using scattering
     !
-    if(check_values) error_msg =  optical_props%validate()
+    error_msg =  optical_props%validate()
```
RobertPincus, line 223: Hi, is this a change you meant to make, @alexeedm ? Now the values of `optical_props` will always be checked for validity.

alexeedm: Yeah, this is just for consistency. `validate()` does checks governed by both `chech_values` and `check_extents`, so when `check_values` is `.false.` you may still want to check the extents.

alexeedm:
@RobertPincus Should be good to merge now!

RobertPincus (approved):
Thanks a lot for the final polishing

rte/kernels-openacc/mo_rte_solver_kernels.F90
naromero77: I think its OK for this PR to be merged, but I still have concerns that this section of code may cause problems in the future for OpenMP offload.

RobertPincus: @naromero77 Feel free to open an issue, a PR, or both. 

Note that we've discussed separating the Jacobian (= optional arguments) kernels for other reasons.

PR #123, opened 2021-06-11
Tang paper related comments are revised

inpolonsky:
I was a bit hectic. I did start from develop


RobertPincus (approved):
Thanks, these are perfectly clear now.

PR #126, opened 2021-07-08
Refactor SW two-stream calculation

RobertPincus:
The initial implementation computed direct and diffuse reflectance and transmittance, then computed the source of diffuse radiation from scattering of the direct beam. This revision combines these calculations in both CPU and GPU kernels to reduce memory use and access.


RobertPincus:
Self-hosted CI is failing because of access issues.

PR #135, opened 2021-08-21
Minor cleanup

MHBalsmeier:
I just resolved three warnings that kept bothering me when compiling:

- unused variables `tsg` and `count` in `extensions/cloud_optics/mo_cloud_optics.F90` removed
- resolving a ` -Wmaybe-uninitialized warning` in the file `rte/kernels/mo_rte_solver_kernels.F90` by moving one statement inside a loop
- typo fixed in README.md

### DIFF
--- a/rte/kernels/mo_rte_solver_kernels.F90
+++ b/rte/kernels/mo_rte_solver_kernels.F90
@@ -871,8 +871,10 @@ contains
         source_dn(i,lay_index) =    Tdir * dir_flux_inc(i)
         dir_flux_trans(i)      = Tnoscat * dir_flux_inc(i)
       end do
+      if (j==nlay) then
+        source_sfc(:) = dir_flux_trans(:)*sfc_albedo(:)
+      endif
     end do
-    source_sfc(:) = dir_flux_trans(:)*sfc_albedo(:)
 
   end subroutine sw_dif_and_source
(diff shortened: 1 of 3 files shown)


rte/kernels/mo_rte_solver_kernels.F90
```diff
@@ -871,8 +871,10 @@ pure subroutine sw_dif_and_source(ncol, nlay, top_at_1, mu0, sfc_albedo, &
         source_dn(i,lay_index) =    Tdir * dir_flux_inc(i)
         dir_flux_trans(i)      = Tnoscat * dir_flux_inc(i)
       end do
+      if (j==nlay) then
```
RobertPincus, line 874: Hi @MHBalsmeier Thanks for these suggestions. This particular suggestion I don't want to adopt because it puts an if-test inside a loop, which may degrade performance quite a bit. We're using a `for` loop precisely because there's a loop carry dependency, so the transmitted direct-beam flux is guaranteed to be correct at the end of the loop.

MHBalsmeier: Yes I assumed this already but the compiler gave a warning when compiled with `-Wall`. I undid it.

RobertPincus (approved):
Good changes all, thanks @MHBalsmeier

PR #137, opened 2021-09-18
Feature: reduce in place

RobertPincus:
In the frequently-used scenario that  broadband flux profiles are requested (using ty_fluxes_broadband) fluxes are accumulated in the solvers, rather than after the fact. On CPUs this reduces memory use.


.github/workflows/self-hosted-ci.yml
```diff
@@ -64,12 +64,15 @@ jobs:
         source compiler_modules
         export RRTMGP_ROOT=$PWD
         export FC=ftn
+        mkdir -p $PWD/tmp
```
RobertPincus, line 67: @alexeedm Can you explain how this temporary directory is used?

examples/all-sky/rrtmgp_allsky.F90
```diff
@@ -338,6 +339,9 @@ program rte_rrtmgp_clouds
   !
   !!$omp parallel do firstprivate(fluxes)
   do iloop = 1, nloops
+    ! Omit the checks starting with the second iteration
+    if (iloop == 2) call rte_config_checks(logical(.false., wl))
```
RobertPincus, line 343: @alexeedm Mmmm, do we want to make this decision on the user's behalf?

alexeedm: This is an example, so all the loops are identical. If the checks pass on the first loop, they do pass on all the rest

RobertPincus: True enough... I guess you're suggesting that timing with the checks off is more useful information than timing with the checks on? Maybe we should note this somewhere (in the output? in the Readme?)

.github/workflows/self-hosted-ci.yml
alexeedm: NVHPC compiler uses temporary folder for compilation. On daint it ran out of space in the default tmp/ folder multiple times, according to the CI logs

examples/all-sky/rrtmgp_allsky.F90
alexeedm: Yep, without the checks we can do timings pretty accurately in this example. Your call regarding the note!

PR #146, opened 2021-11-12
Top at 1 fix

alexeedm:


RobertPincus (approved):
Thanks, we'll think of this as a temporary fix before making top_at_1 a variable associated with the class.

PR #148, opened 2021-11-30
Feature: reordered gas optics

RobertPincus:
Work by Erik Pfister at DKRZ eliminates the need for the transposition of large arrays. Further refinements by Dmitry Alexeev of Nvidia fusing loops and otherwise accelerating the OpenACC gas optics kernels.


RobertPincus:
Note that OpenMP GPU compilation fails on CCE 11 and 12 by causing an internal compiler error.

PR #151, opened 2021-12-14
Simplify index logic in subroutine interpolate_3D in mo_gas_optics_kernels.F90

climbfuji:
This PR simplifies the logic in subroutine `interpolate_3D` in `mo_gas_optics_kernels.F90`. I have not run any tests with this except for compiling it, but from a visual inspection this should produce the same results.

PR #152, opened 2021-12-20
For develop: Simplify index logic in subroutine interpolate_3D in mo_gas_optics_kernels.F90

climbfuji:
This PR replaces https://github.com/earth-system-radiation/rte-rrtmgp/pull/151. It simplifies the logic in subroutine `interpolate_3D` in `mo_gas_optics_kernels.F90`. I have not run any tests with this except for compiling it, but from a visual inspection this should produce the same results.


RobertPincus (approved):
Thanks, this is very clean.

PR #153, opened 2022-01-20
make data download agnostic to certificate issues

Try2Code:
hi!

I experience problems with this data download because wget usually does test for bad certificates and avoids the download. This is my workaround.


Try2Code:
I realized that there is 2nd download related script written python. Another option could be remove the shell script and replace it with this python version

RobertPincus:
@Try2Code Thanks for this. Are you able to change the target of the PR to the "develop" branch?

Try2Code:
created a new on based on 'develop' #154

RobertPincus (commented):
Seems great, thanks for the suggestion.

PR #154, opened 2022-01-28
make data download agnostic to certificate issues

Try2Code:
now on top of develop branch (as requested in #153)


Try2Code:
hi!
is there anything I can do? the error does not seem to be related to my change ...

RobertPincus:
@Try2Code One of the failures is expected, but the widespread failures starting with gfortran-9 may be related to the same expired certificates. I'll ask about getting the certificates updated, which would make the problem go away, and run checks locally, but I'm buried with other tasks and this might be a week.

Try2Code:
I would try to get rid of the certificate check - (a) this is something you have under control and (b) you won't  rely on other ppl in the future on that topic

... just saying

RobertPincus:
@Try2Code Just to let you know - I'm not ignoring you, just killed with teaching, will attend to this in March.

Try2Code:
all cool @RobertPincus  - thx for letting me know

RobertPincus:
@Try2Code If we move from `wget` to downloading files with Python, as in #169, would this resolve the issue?

PR #155, opened 2022-01-28
Fix comment with compile flags

mathomp4:
As found by @dr0cloud, it looks like a bad copy-paste happened. So we fix it up. 😄 

I didn't make an issue for this since it's so minor, but if you'd like me to for record-keeping, let me know.


RobertPincus (approved):
Doh! Thanks a lot.

PR #157, opened 2022-02-04
Modest mo_cloud_sampling changes (cosmetic, commentary, speed, and possible bug fix to exp_ran)

dr0cloud:
Hi @RobertPincus, thanks to @mathomp4, we are finally synced with your repository and able to offer PRs etc. Here is a modest one to fix/improve the existing cloud sampling. My next goal is to add some greater flexibility/capability to the cloud subcolumn generation/overlap system. Peter
 
new line#:
94,226,230,234,239,243: improved/corrected error message.
126,173,189,289: cosmetic or more consistent with elsewhere in code.
170&260: execution for cloud_lay_fst redundant ... overwritten subsequently.
195-203: more detailed commentary.
284-285: this looks like a bugfix ... existing 1.5 code would potentially leave cloud_mask for intra-cloud clear layers uninitialized.

### DIFF
--- a/extensions/cloud_optics/mo_cloud_sampling.F90
+++ b/extensions/cloud_optics/mo_cloud_sampling.F90
@@ -91,7 +91,7 @@ contains
     nbnd = clouds%get_nband()
     ngpt = clouds_sampled%get_ngpt()
     if (any([size(cloud_mask,1), size(cloud_mask,2), size(cloud_mask,3)] /= [ncol,nlay,ngpt])) then
-      error_msg = "draw_samples: cloud mask and cloud optical properties have different ncol and/or nlay"
+      error_msg = "draw_samples: cloud mask and cloud optical properties have different ncol, nlay and/or ngpt"
       return
     end if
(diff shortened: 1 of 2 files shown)


dr0cloud:
Note: "exp_ran" is a somewhat unfortunate naming since the method as coded can handle the case of negative inter-layer correlations, which are not possible for pure exponential correlation decay with layer separation. I would personally prefer
this method to be renamed "gen_ran" (and "exponential" -> "generalized" in the comments), especially since I have other
purely exponential overlap methods (which only model the transition from random to maximal overlap) that I would like to add to this module. But I understand that would break backward compatibility. Perhaps thats OK if its only a name 
change and if the new methods I add do not have exactly the old name? Peter.

dr0cloud:
Hi @RobertPincus , no problem at all ... and please don't mind my previous comment about the exp_ran name too much ... its more debatable. By March, I hope the have a slightly more weighty "PR" that suggests a way to generalize the current cloud sampling for including condensate overlap as well ... I'll submit it as a PR, but it will really be more of a suggestion and a good starting place for a discussion. Peter.

RobertPincus (approved):
Thanks for these, @dr0cloud

PR #163, opened 2022-03-31
Adding documentation capabilities

Gfacco:
This PR integrates the building of a user documentation site using Jekyll, as well as a developer documentation site wich uses FORD to create an automated documentation by parsing the code directly


RobertPincus:
@Gfacco Thanks very much for this. Two requests. First, can you merge the last commit on `develop` into this branch so it's up to date? (Only one file has changed, for the self-hosted CI.) Second, I still find it hard to believe we have to add the whole Ruby(?) bundle of > 5K files. Seems like, if a single file like the top-level index gets overwritten when you initialize Ruby, we could save the file we need under a different name and move it into position after Ruby is initialized in the workflow.

RobertPincus:
@Gfacco That's indeed much cleaner, thanks. It seems some of the source files have been changed  - there are even some unresolved merge strings in there. Seems like the only source file changes should be related to comments that FORD can read. Do you know what's up?

Gfacco:
@RobertPincus i resolved the merge conflict errors, im not sure what happened there.  We had added the FORD flags to some of the comments in some files in order to show how FORD parses them.  These can be removed if you prefer.

RobertPincus (approved):
Thanks so much for this, @Gfacco

PR #164, opened 2022-04-06
Add labels to "end type" statements to eliminate warnings in CCPP pre…

mkavulich:
This PR adds labels for "end type" statements in the rrtmgp code. These end statements are technically not required by the CCPP standards, but all other "end" statements in CCPP must be labeled, and this helps reduce the potential for ambiguity in the code. These changes will also eliminate warnings when running ccpp_prebuild.py and the new ccpp_track_variables.py script (https://github.com/NCAR/ccpp-framework/pull/419):

```
WARNING:root:In file mo_gas_optics_rrtmgp.F90, module mo_gas_optics_rrtmgp
WARNING:root:Encountered closing statement "end type" without type name; assume type_name is ty_gas_optics_rrtmgp
WARNING:root:In file mo_optical_props.F90, module mo_optical_props
WARNING:root:Encountered closing statement "end type" without type name; assume type_name is ty_optical_props
WARNING:root:In file mo_optical_props.F90, module mo_optical_props
```

etc.

This should have no change in results, as it is simply for human (and CCPP) readability.

Issue #161, opened 2022-03-30
Memory layout in gas optics kernels


RobertPincus:
An issue received via email: 

```
From: Thomas Jahns <[jahns@dkrz.de](mailto:jahns@dkrz.de)>
Sent: Monday, March 7, 2022 9:57 AM
To: [rrtmgp@aer.com](mailto:rrtmgp@aer.com) <[rrtmgp@aer.com](mailto:rrtmgp@aer.com)>
Cc: Claudia Frauen <[frauen@dkrz.de](mailto:frauen@dkrz.de)>; Jörg Behrens <[behrens@dkrz.de](mailto:behrens@dkrz.de)>
Subject: Memory layout in rrtmgp/kernels/mo_gas_optics_kernels.F90
 
Hi,

I was roped in by some colleagues to look at some issues in the mo_gas_optics_kernels module and noticed two other, independent problems for compiler optimization of the module:

1. all routines are made public and bind(c) which prevents inline for compilers adhering to strict semantics for symbols that can be interpositioned. Would it be possible to change the declaration of interpolate3D_byflav to private and remove the bind(c) to permit Fortran assumed shape arguments.

2. the above would allow to change the fields from e.g.

integer, dimension(2, ncol,nlay,nflav), intent(out) :: jeta
real(wp), dimension(2, ncol,nlay,nflav), intent(out) :: col_mix
real(wp), dimension(2,2,2,ncol,nlay,nflav), intent(out) :: fmajor
real(wp), dimension(2,2, ncol,nlay,nflav), intent(out) :: fminor

to

integer, dimension(ncol,2, nlay,nflav), intent(out) :: jeta
real(wp), dimension(ncol,2, nlay,nflav), intent(out) :: col_mix
real(wp), dimension(ncol,2,2,2,nlay,nflav), intent(out) :: fmajor
real(wp), dimension(ncol,2,2, nlay,nflav), intent(out) :: fminor

and enable the compiler to process the data with multiple packed SIMD-streams instead of emitting instructions for vector gather/scatter operations of data that is essentially in AoS layout.

I think I could submit a corresponding code change if the idea seems possible in general.

Kind regards, Thomas
```

RobertPincus:
An initial response by email: 
```
 It should be possible to remove C bindings from any routines that aren’t called by the Fortran front-ends. I’m skeptical, though, that using assumed-shape will lead to better performance than using assumed-size. 
```

RobertPincus:
[...]

Issue #158, opened 2022-02-15
Generic names in C bindings


skosukhin:
Currently, certain subroutines are exposed via C bindings under very generic names, e.g. [`interpolation`](https://github.com/earth-system-radiation/rte-rrtmgp/blob/e9b7cbd70f286e6152e69f52c1a80018a13913b8/rrtmgp/kernels/mo_gas_optics_kernels.F90#L34). That might trigger a name collision in a user application.

A possible solution is to come up with some common prefix and use it for all public functions of the library, e.g. `rrtm_interpolation`.

RobertPincus:
@skosukhin Thanks, this is a good idea. I'll try to implement in March some time.

PR #165, opened 2022-04-09
Restrict visibility, end type labels; fix Python in CI

RobertPincus:
1. Narrow the range of public procedures in the Fortran kernel modules; the remaining ~12 are the kernel API. 
2. Remove C bindings from any non-public functions (#161)
3. Add labels to end type statements (#164)
4. Update use of conda in continuous integration using Github Actions


RobertPincus:
Also 5. Remove unneeded array-reordering kernels

RobertPincus:
And 6. C-bindings are now prefixed with "rte_" or "rrtmgp_"  (#158)

PR #167, opened 2022-04-11
Update Scripts

everythingfunctional:
This makes building the documentation slightly more convenient


RobertPincus (approved):
Nice streamlining, thanks a lot.

PR #169, opened 2022-05-07
Updated testing environments

RobertPincus:
Use `environment.yml` file to detail Python packages needed for testing and validation/verification, to be obtained from `conda`. The `environment-dev.yml` variant also includes gfortran compilers and the netCDF stack and can be used for development. 

Move to Python for downloading test files - should resolve #154 by eliminating `wget`

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -113,7 +113,11 @@ jobs:
       uses: conda-incubator/setup-miniconda@v2
       with:
         miniforge-version: latest
-    - name: Install python packages
+        activate-environment: rte_rrtmgp_test
+        environment-file: environment.yml
+        python-version: 3.9
+        auto-activate-base: false
+  - name: Install python packages
       run: conda install --yes urllib3 netcdf4 xarray dask numpy scipy matplotlib seaborn colorcet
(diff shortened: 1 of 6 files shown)


RobertPincus:
@d70-t A question for you, since you understand the big picture so well. This PR makes explicit the Python packages used in testing and validation and provides an `environment.yml` file to install these within conda. The CI works fine on Linux VMs but a local installation on my ARM Mac fails because one of the C libraries pointed to by the current version of `numpy` is missing.  It's not a numpy problem, really, but a packaging/installation problem - where would I go for support?

d70-t:
It seems like I can't see any log output from the failing checks (seems like a permissions error). Otherwise, I doubt that I can be of much help here. I don't have an ARM Mac at hand and also am not very well informed about the conda installation process.

I'd try to run the installation steps manually on the target machine (or a similar one) and try to figure out where it hangs... It seems like the `numpy` packages you are fetching have pre-compiled binaries shipped (which makes installation fast) but these are probably not for ARM (which makes them not work...). You can try to find a way of installing them without binaries (in pip land, it would be something like `pip install --no-binary :all:`) or you can try to approach the packaging people and ask for their help.

PR #171, opened 2022-05-22
feature/single precision

climbfuji:
### WORK IN PROGRESS

- Switch working precision to single precision of preprocessor flag `-DSINGLE_PRECISION` is set

### DIFF
--- a/rte/mo_rte_kind.F90
+++ b/rte/mo_rte_kind.F90
@@ -26,7 +26,11 @@ module mo_rte_kind
   !
   ! Floating point working precision
   !
+#ifdef SINGLE_PREC
+  integer, parameter :: wp = sp
+#else
   integer, parameter :: wp = dp
+#endif
 
   !
   ! Logical - for use with kernels


RobertPincus:
@climbfuji Note the same functionality is already on the `develop` branch in commit 345aefb72d4c38deb7f5a353777c479904bf1d13. 

I like to move things from `develop` to `main` and then out to `autoconf` and `ccpp`. If this is pressing please let me know and I'll update `develop`. 

Note also that this won't work - some calculations will fail if done in single precision. This is something we're working on and can prioritize if you have an immediate need.

climbfuji:
> @climbfuji Note the same functionality is already on the `develop` branch in commit [345aefb](https://github.com/earth-system-radiation/rte-rrtmgp/commit/345aefb72d4c38deb7f5a353777c479904bf1d13).
> 
> I like to move things from `develop` to `main` and then out to `autoconf` and `ccpp`. If this is pressing please let me know and I'll update `main`.
> 
> Note also that this won't work - some calculations will fail if done in single precision. This is something we're working on and can prioritize if you have an immediate need.

@RobertPincus Absolutely! You'll be receiving an email in a second that provides more background.

Chiil:
Would it be a good idea to add a prefix `RTE_RRTMGP_` or something similar to preprocessor flags?

RobertPincus:
Closed with a4fe30cf4dab2e5fd8d3ab6f11683a82ae584475 but note that code compiled as single precision almost certainly won't work.

PR #176, opened 2022-06-03
Fixing documentation workflow issues

Gfacco:
This should fix the issues where the link to the github pages site was not working.

A small typo was introduced where instead of the contents of the `_site` folder being moved to the `public` folder, the whole folder was moved. By changing the copy command as seen below will result in the desired behaviour.

`cp -a _site ../../public` 
since public already exists this moves the contents of `_site` to `public/_site`

`cp -a _site/* ../../public` 
this will move only the contents of `_site` to `public`


RobertPincus (approved):
Thanks, happy to try this.

PR #180, opened 2022-06-24
32-bit physics with FV3_RAP

SamuelTrahanNOAA:
This adds support for 32-bit physics to the FV3, based on prior work on the Neptune model. The 64-bit physics should not change results. The only change in this repository is to add a flag to enable single precision. The code for that was copied from the authoritative branch.

See the top-level PR here for details:

https://github.com/ufs-community/ufs-weather-model/pull/1215

The issue for this is here:

https://github.com/ufs-community/ufs-weather-model/issues/1288

### DIFF
--- a/rte/mo_rte_kind.F90
+++ b/rte/mo_rte_kind.F90
@@ -26,7 +26,11 @@ module mo_rte_kind
   !
   ! Floating point working precision
   !
+#ifdef RTE_USE_SP
+  integer, parameter :: wp = sp
+#else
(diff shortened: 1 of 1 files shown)


RobertPincus:
@SamuelTrahanNOAA The `dtc/ccpp` branch has been maintained by @dustinswales. We could incorporate this single change but it would sure be nicer to have the branch updated with all changes to `main`. I dunno if Dustin has time to manage this, or what the flow should be since this branch is specific to UFS (CCPP)/

PR #181, opened 2022-06-29
Xyuan/rrtmgp hip

xyuan:
fix the issue that HIP kernel can't call host function.


xyuan:
This branch looks pretty strange, so I close this PR

PR #182, opened 2022-07-01
Fix Documentation CI

everythingfunctional:
Upgrading to newer versions of things fixes it.


RobertPincus (approved):
Simple enough - though hard to see how changing the gfortran version would matter. Thanks a lot.

PR #183, opened 2022-08-01
fix(doc/build_documentation.sh): missing mkdir

everythingfunctional:


RobertPincus (approved):
Doh! Thanks a lot, and egg on my face for having missed this difference.

PR #188, opened 2022-10-05
ongoing work to incorporate CCE 14.0 OpenACC-capable version

vectorflux:
ongoing work to incorporate CCE 14.0 OpenACC-capable version.


vectorflux:
@alexeedm : I could use your help.   There are slight OpenACC differences in 4 files in order to get this to compile with CCE 14.0.3. But the differences in rrtmgp/mo_gas_optics_rrtmgp.F90 break nvidia_gpu with the error:

FATAL ERROR: variable in data clause is partially present on the device: name=optical_props
 file:/scratch/snx3000/rpincus/daint102-work/rte-rrtmgp/rte-rrtmgp/build/../rrtmgp/mo_gas_optics_rrtmgp.F90 combine_abs_and_rayleigh line:1922

Note I've put !$acc data copyin(optical_props) outside of the enter/exit data regions of its constituents.  I literally cannot see the problem with this.  I'd be indebted to you if you could take a look.

alexeedm:
@vectorflux should be fixed. You can't allocate base class in nvhpc compiler and use it as derived in the later present statements. I'm checking with Jeff Larkin if that's standard-conforming, and will submit a bug if it's not

vectorflux:
@alexeedm Thanks so much for the assistance.  I definitely see the issue and understand the fix, which essentially goes back to the original code.  Unfortunately the code now crashes the CCE-14 compiler, no doubt a bug there.   I'll report it to John Levesque, but have to find a solution which makes both compilers happy.  Thanks also for asking Jeff if Nvidia is standard-compliant.

vectorflux:
I had to solve this with a very ugly code divergence.   If @alexeedm talks to Jeff and I talk to John, hopefully this can be removed later.   There are a few more small fixes needed, then I hope this PR can go forward (albeit the problem with Cray OMP offloading needs to be fixed).

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -1516,10 +1539,12 @@ function compute_optimal_angles(this, optical_props, optimal_angles) result(err_
     integer  :: ncol, nlay, ngpt
     integer  :: icol, ilay, igpt, bnd
     real(wp) :: t, trans_total
+    integer, allocatable :: bands(:)
```
RobertPincus, line 1542: Maybe `integer :: bands(optical_props%get_ngpt())` to avoid explicit allocation ?

RobertPincus (changes requested):
The CI at CSCS is using the default CCE module. Is CCE 14 available on daint? If so, do you want to change `.github/workflows/self-hosted-ci.yml' to point to it?

fomics:
@RobertPincus I think I have the fix for CCE 14.0.0, but once again I can pull but not push to the develop_cce_gpu branch.  It might be related to .ssh keys, but if so one would think that the pull would not even work.

fomics:
[self-hosted-ci.yml.txt](https://github.com/earth-system-radiation/rte-rrtmgp/files/9888502/self-hosted-ci.yml.txt)

RobertPincus:
@vectorflux Sorry to be slow; Fridays are a mess for me. I've updated the self-hosted CI as per your attached file. I'm surprised you can't commit and agree this is likely to be an issue with ssh keys. If you can't resolve the key question maybe you want to branch to a personal repo? I'm also happy to commit changes on your behalf.

RobertPincus:
@vectorflux Any chance of getting this merged in the coming days?

vectorflux:
sorry if I dropped the ball on this one.  Let's see if I can fix the SSH keys.

vectorflux:
keys now working again.  Have to go cook a chicken (not turkey) for Thanksgiving -- hopefully I can finish this on the side.

vectorflux:
Container tests failing with:

KeyError: [<class 'netCDF4._netCDF4.Dataset'>, ('http://esgf-data1.llnl.gov/thredds/dodsC/css03_data/CMIP6/RFMIP/AER/LBLRTM-12-8/rad-irf/r1i1p1f1/Efx/rsd/gn/v20190514/rsd_Efx_LBLRTM-12-8_rad-irf_r1i1p1f1_gn.nc',), 'r', (('clobber', True), ('diskless', False), ('format', 'NETCDF4'), ('persist', False)), 'd8ed229a-40cd-41b8-95b2-21813ef6371c'

Any ideas?

RobertPincus:
@vectorflux I've poked at this a little over the Thanksgiving break. I admit to being mystified. The facility providing access to the verification data can be a little flakey, but this doesn't explain why the containerized CI is failing and the the other pathways (e.g. normal CI, or self-hosted CI) are not. 

The containers have been rebuilt recently, for hat it's worth

RobertPincus:
@vectorflux Can you add 
`      fail-fast: false`
to the `strategy` section of `.github/workflows/containerized-ci.yml'? The failure to download the file is happening in a part of the CI that's only being exercised here (with the `ifort` compiler)

vectorflux:
I'm trying to reproduce manually what self-hosted-ci is doing, but am struggling.   Here some comments:

- The CCE compiler information is not documented in Compiler-flags.md

-  export FCFLAGS="-hnoacc -homp -O0" is not wise, because -O0 might disable acceleration

-  export RTE_KERNELS="openacc" for the OpenMP-acc mode is unaesthetic.  "accelerator" would be better.

- With CCE 12.0.3 (Daint default) it crashes with an internal compiler error in rrtmgp/mo_gas_optics_rrtmgp.F90. 

- With CCE 14.0.0 (latest on Daint) rrtmgp/mo_gas_optics_rrtmgp.F90 has the following compilation error:  ftn-7032 crayftn: ERROR COMPUTE_OPTIMAL_ANGLES, File = ../rrtmgp/mo_gas_optics_rrtmgp.F90, Line = 1550 Unsupported OpenMP construct Calls

-  It is not yet clear to me how self-hosted-ci.yml (which appears to use 14.0.0) gets through to the link stage.  But my suspicion is that the linker (which must call nvlink) needs some additional flag for OpenMP-acc.

rrtmgp/mo_gas_optics_rrtmgp.F90
```diff
@@ -1513,10 +1537,12 @@ function compute_optimal_angles(this, optical_props, optimal_angles) result(err_
     integer  :: ncol, nlay, ngpt
     integer  :: icol, ilay, igpt, bnd
     real(wp) :: t, trans_total
+    integer, allocatable  :: bands(:)
```
RobertPincus, line 1540: @vectorflux Why not use `integer :: bands(optical_props%get_ngpt))` ? Saves explicit de/allocation below.

rrtmgp/mo_gas_optics_rrtmgp.F90
vectorflux: `integer :: bands(optical_props%get_ngpt())` was failing with the CCE compiler, that is why it has to be allocated explicitly, at least for the time being.

rrtmgp/mo_gas_optics_rrtmgp.F90
vectorflux: `integer :: bands(optical_props%get_ngpt))` fails with the CCE 14.0.3 compiler, so it needs to be allocated explicitly, at least for the time being.

skosukhin: @vectorflux it might make sense to document the workaround with
```fortran
#if defined _CRAYFTN && _REALEASE_MAJOR == 14 && _RELEASE_MINOR == 0 && _RELEASE_PATCHLEVEL == 3
# define CRAY_WORKAROUND
#endif

#ifdef CRAY_WORKAROUND
integer, allocatable  :: bands(:)
#else
integer :: bands(optical_props%get_ngpt))
#endif
```

m214089:
There is a typo in the above snippet: 

_REALEASE_MAJOR 

take care!

skosukhin:
@m214089 Good catch. I've just copied and extended https://gitlab.dkrz.de/dkrz-sw/yaxt/-/blob/69e5ae4c097372ca8eefb683b3f97548234bc5a3/tests/test_misc_f.f90#L141. @tjahns, you might want to fix it in your code (or was there a typo in the name of the macro in the old version of Cray compiler?).

PR #189, opened 2022-11-03
Unused local variables removed

MHBalsmeier:
Hi, I removed some unused local variables.

There are some more possibilities for a further clean-up:

- The functions `interpolate2D` and `interpolate3D` are unused. Can they be removed or are they meant to evolve?
- There are quite a few unused dummy arguments. Can they all be removed from the function calls or are they still there for compatibility reasons? Example: Dummy argument `gas_names` in subroutine `reduce_minor_arrays`.

### DIFF
--- a/rrtmgp/kernels/mo_gas_optics_kernels.F90
+++ b/rrtmgp/kernels/mo_gas_optics_kernels.F90
@@ -527,7 +527,7 @@ contains
     ! -----------------
     ! local variables
     real(wp) :: k(ngpt) ! rayleigh scattering coefficient
-    integer  :: icol, ilay, iflav, ibnd, igpt, gptS, gptE
+    integer  :: icol, ilay, iflav, ibnd, gptS, gptE
     integer  :: itropo
     ! -----------------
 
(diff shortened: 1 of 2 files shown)


rrtmgp/kernels/mo_gas_optics_kernels.F90
```diff
@@ -527,7 +527,7 @@ subroutine compute_tau_rayleigh(ncol,nlay,nbnd,ngpt,         &
     ! -----------------
     ! local variables
     real(wp) :: k(ngpt) ! rayleigh scattering coefficient
-    integer  :: icol, ilay, iflav, ibnd, igpt, gptS, gptE
```
RobertPincus, line 530: Thanks @MHBalsmeier these are fine changes to make

RobertPincus (approved):

RobertPincus:
@MHBalsmeier Thanks for the minor cleanups. In principle there's no problem removing unused arguments, procedures, or other codes, as long as changes continue to pass the continuous integration with both `default` and `openacc` sets of kernes. We haven't worried so much about the initialization functions like `reduce_minor_arrays` because they're called once and not visible outside the module but you're welcome to prune them.

PR #190, opened 2022-11-04
Unused functions and dummy arguments removed

MHBalsmeier:
As mentioned in #189, I removed some unused functions and dummy arguments.


MHBalsmeier:
Some unused dummy arguments I tried to remove are actually needed by the OpenACC versions, so I had to undo one commit. I'll leave it at that for now.

RobertPincus (approved):
Thanks, these are useful simplifications.

PR #192, opened 2022-12-09
default(none) fixes for CCE

jussienko:


vectorflux (approved):
These changes are fine.  We should revisit this when DEFAULT(PRESENT) in DATA clauses is supported

PR #200, opened 2022-12-23
Update CI

skosukhin:
1. Fixes caching of RFMIP files. Currently, the files are cached but not used: each job still downloads the files.
3. Installs `zstd` on top of the container images, which makes `CI` and `Containerized-CI` use the same cache (see [here](https://github.com/actions/cache#cache-version) why).
3. Introduces caching of the Conda environment. This saves ~500MB of incoming traffic per job (~4x500MB per workflow run).
4. Removes entries from `environment.yml` that are not needed to run the CI jobs. This reduces the cache file from ~500MB to ~280MB. I guess we can reduce it further if we switch to `apt-get install` and `pip install`.
5. Replaces custom installation of `HDF5`, `NetCDF-C` and `NetCDF-Fortran` in the `CI` jobs with Ubuntu package `libnetcdff-dev`, which works because Fortran module files for Gfortran 9 and Gfortran 10 are compatible.
6. Drops installation of Gfortran in `doc-deployment.yml` because it seems redundant to me but I might be wrong: I haven't thoroughly compared the results before and after the change.
7. Makes the environment module manipulations on Daint (`self-hosted-ci.yml`) easier (hopefully). The exact mechanism is used in ICON development.
8. Switches to a newly installed Python environment on Daint, which saves cumbersome environment manipulations before we can start using it.
9. Introduces general refactoring of the CI configurations: they look similar to each other as much as possible, which is supposed to make it easier to understand what we test.

Tip: it might happen at some point that the cached files (i.e. RFMIP or Conda) do not match changes in the code and the workflows will start failing because of that. Some time ago, it was not possible to update/remove the cache files. The only way to tell the jobs that they should not use the existing cache but download the files on their own was to change the `key`. Now, we can easily remove invalid/obsolete caches [here](https://github.com/earth-system-radiation/rte-rrtmgp/actions/caches).

Recommendations on the containers:
1. Intel container should source `/opt/intel/oneapi/setvars.sh` (e.g. in `~/.profile`) so that the respective CI jobs don't have to do that over and over again.
2. The containers should have `zstd` installed for the same reason.
3. ~It makes sense to reduce the size of the containers by appending `&& rm -rf /var/lib/apt/lists/*` to the last `apt-get install` command of a container.~ (the reduction in size is negligible)
4. ~It's a good general practice to create a non-root user in the container and switch to it. In my opinion, it makes sense to adopt that.~ (see https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user)

### DIFF
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -25,7 +25,6 @@ jobs:
         - fortran-compiler: nvfortran
           fcflags: "-Mallocatable=03 -Mstandard -Mbounds -Mchkptr -Kieee -Mchkstk"
           image: "earthsystemradiation/rte-rrtmgp-ci:nvfortran"
-        - netcdf-fortran: /opt/netcdf-fortran
     container:
       image: ${{ matrix.image }}
     env:
@@ -34,7 +33,7 @@ jobs:
       FCFLAGS: ${{ matrix.fcflags }}
       # Make variables:
       NCHOME: /dummy
-      NFHOME: ${{ matrix.netcdf-fortran }}
+      NFHOME: /opt/netcdf-fortran
       RRTMGP_ROOT: ${{ github.workspace }}
       RTE_KERNELS: ${{ matrix.rte-kernels }}
       RUN_CMD:
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -20,10 +20,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        fortran-compiler: [ifort, nvfortran]
+        fortran-compiler: [ifort, ifx, nvfortran]
         rte-kernels: [default, openacc]
         include:
         - fortran-compiler: ifort
+          fcflags: "-m64 -g  -traceback -heap-arrays -assume realloc_lhs -extend-source 132 -check bounds,uninit,pointers,stack -stand f08"
+          image: "earthsystemradiation/rte-rrtmgp-ci:ifort"
+        - fortran-compiler: ifx
           fcflags: "-m64 -g  -traceback -heap-arrays -assume realloc_lhs -extend-source 132 -check bounds,uninit,pointers,stack -stand f08"
           image: "earthsystemradiation/rte-rrtmgp-ci:ifort"
         - fortran-compiler: nvfortran
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -9,7 +9,10 @@ on:
 
 defaults:
   run:
-    shell: bash
+    # Enable compilers by using login shell:
+    #   Contrary to what the documentation says, GitHub Actions override the
+    #   entrypoint: https://github.com/actions/runner/issues/1964
+    shell: bash -leo pipefail {0}
 
 jobs:
   Containerized-CI:
@@ -26,8 +29,7 @@ jobs:
         - fortran-compiler: nvfortran
           fcflags: "-Mallocatable=03 -Mstandard -Mbounds -Mchkptr -Kieee -Mchkstk"
           image: "earthsystemradiation/rte-rrtmgp-ci:nvfortran"
-        - netcdf-c: /home/runner/netcdf-c
-        - netcdf-fortran: /home/runner/netcdf-fortran
+        - netcdf-fortran: /opt/netcdf-fortran
     container:
       image: ${{ matrix.image }}
     env:
@@ -35,7 +37,7 @@ jobs:
       FC: ${{ matrix.fortran-compiler }}
       FCFLAGS: ${{ matrix.fcflags }}
       # Make variables:
-      NCHOME: ${{ matrix.netcdf-c }}
+      NCHOME: /dummy
       NFHOME: ${{ matrix.netcdf-fortran }}
       RRTMGP_ROOT: ${{ github.workspace }}
       RTE_KERNELS: ${{ matrix.rte-kernels }}
@@ -47,16 +49,6 @@ jobs:
     # Checks-out repository under $GITHUB_WORKSPACE
     #
     - uses: actions/checkout@v3
-    #
-    # Install Zstd
-    #
-    - name: Install Zstd
-      run: |
-        # We skip 'apt-get update' to keep this step lightweight. However, that
-        # might lead to a failure, which we tolerate because the only reason we
-        # run it is to make cache files compatible across workflows.
-        # TODO: update the containers so that we could skip the following:
-        apt-get install zstd || true
     #
     # Cache RFMIP files
     #
@@ -80,14 +72,6 @@ jobs:
             cp "${file}" "${GITHUB_WORKSPACE}/examples/rfmip-clear-sky/"
           fi
         done
-    #
-    # Finalize build environment
-    #
-    - name: Finalize build environment
-      if: contains('ifort ifx', matrix.fortran-compiler)
-      run: |
-        # TODO: update the Intel container so that we do not have to the following:
-        echo "BASH_ENV=/opt/intel/oneapi/setvars.sh" >> "${GITHUB_ENV}"
     #
     # Build libraries, examples and tests
     #
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -1,11 +1,15 @@
 name: Continuous integration in a box
-on: 
-  push: 
-    branches-ignore:    
-      - documentation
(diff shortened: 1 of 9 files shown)


.github/workflows/containerized-ci.yml
```diff
@@ -1,84 +1,123 @@
...
+    #
+    - name: Install Zstd
       run: |
-        source /opt/intel/oneapi/setvars.sh || true
-        cd ${RRTMGP_ROOT}
-        ${FC} --version
-        make libs
-        make -C build separate-libs
-    ############################################################################
+        # We skip 'apt-get update' to keep this step lightweight. However, that
```
RobertPincus, line 55: @skosukhin The containers have been updated to install `zstd`. Can this step be removed?

.github/workflows/containerized-ci.yml
```diff
@@ -1,84 +1,123 @@
...
-        source /opt/intel/oneapi/setvars.sh || true
-        export LD_LIBRARY_PATH=${NFHOME}/lib:${LD_LIBRARY_PATH}
-        make tests
-    - name: Comparison
-      run: make check
-    ############################################################################
-    - name: Validation plots
-      if: contains(matrix.container, 'ifort') && contains(matrix.rte-kernels, 'default')
+        # TODO: update the Intel container so that we do not have to the following:
+        echo "BASH_ENV=/opt/intel/oneapi/setvars.sh" >> "${GITHUB_ENV}"
```
RobertPincus, line 90: @skosukhin Can you please explain precisely what should be done in the Intel container to be able to remove this step?

environment.yml
```diff
@@ -5,12 +5,6 @@ name: rte_rrtmgp_test
 
 dependencies:
   - conda-forge::python=3.9
-  - conda-forge::urllib3
   - conda-forge::netcdf4
   - conda-forge::xarray
   - conda-forge::dask
-  - conda-forge::numpy
```
RobertPincus, line 12: @skosukhin I understand that removing numpy etc. will streamline the CI environment, but then we can't make the validation plots in `tests` which we do with the `ifort` image. Could we instead create an `environment-noplots.yml`? Also please leave `numpy` as a dependency - it follows from `dask` and `xarray` but I'd rather be explicit

.github/workflows/containerized-ci.yml
```diff
@@ -1,84 +1,123 @@
...
+      image: ${{ matrix.image }}
     env:
-      NCHOME: /home/runner/netcdf-c
-      NFHOME: /home/runner/netcdf-fortran
-      RFMIP_DIR: /home/runner/rfmip-files
+      # Core variables:
+      FC: ${{ matrix.fortran-compiler }}
+      FCFLAGS: ${{ matrix.fcflags }}
+      # Make variables:
+      NCHOME: ${{ matrix.netcdf-c }}
```
RobertPincus, line 38: @skosukhin Can you explain why it's preferable to use environment variables from the `matrix` rather than setting them within the job?

.github/workflows/continuous-integration.yml
```diff
@@ -1,150 +1,120 @@
...
-      run: |
-        sudo apt-get install gfortran-10 gcc-10
-        echo "CC=gcc-10" >> $GITHUB_ENV
-
-    ############################################################################
+    # Install NetCDF-Fortran (compatible with both compilers)
     #
-    # Netcdf C and Fortran
+    - name: Install NetCDF-Fortran
+      run: sudo apt-get install libnetcdff-dev
```
RobertPincus, line 58: @skosukhin I love this idea. Will it also work if we move to Ubuntu 22.04 and use gfortran 10, 11, 12? Should I install netcdff-dev this way in the containers for uniformity?

.github/workflows/module_switcher
```diff
@@ -0,0 +1,116 @@
+switch_for_module ()
```
RobertPincus, line 1: @skosukhin I do not understand what this module is doing. I understood that this is taken from another CI instance running at CSCS? Could you add a line or three of commentary to explain this?

RobertPincus:
@skosukhin This is wonderful, thanks very much. On the containers, if you're able to open a pull request against https://github.com/earth-system-radiation/containers I'm happy to consider it. I can make a branch for you to target if you want.

.github/workflows/continuous-integration.yml
RobertPincus: @skosukhin Scratch the suggestion about containers - of course netCDF Fortran has to be built with the Fortran compiler we're testing.

RobertPincus:
@skosukhin You'll see that everything now runs on Ubunutu 22.04 and that I've followed your lead and use `libnetcdff-dev` installed via `apt-get` in the `gfortran` testing via Github actions

skosukhin:
This now depends on https://github.com/earth-system-radiation/containers/pull/9.

.github/workflows/containerized-ci.yml
skosukhin: It could be something like `RUN echo ". /opt/intel/oneapi/setvars.sh" >> /etc/profile.d/oneapi.sh` in the Dockerfile but this is not needed after switching to the official Intel image.

environment.yml
skosukhin: The `ifort` image uses `pip` to install the plotting packages, so it would not be affected by these changes. I also thought that `environment-dev.yml` was supposed to deliver the full development environment. It looks like I misinterpreted the comment in `environment.yml`. Nonetheless, I followed your suggestion and added `environment-noplots.yml`. And yes, my bad, `numpy` should be explicitly listed since it is used directly (i.e. not via calls to some other package) in the check scripts.

.github/workflows/containerized-ci.yml
skosukhin: I'm sorry, I see two questions here:

1. Why do I set environment variables for all steps of a job and not only for steps that really need them?
    
    I simply prefer having all environment variables set in one place. Of course, a counter-argument to that is that we might miss certain side effects for steps that should not need the variables but implicitly use them.

2. Why do I set variables in the matrix and then re-use them as is to set environment variables?
    
    I consider variables in the matrix as input parameters for a job and environment variables as something that enables it. For example, the build system in `develop` is designed and tested with the requirement that `FCFLAGS`, `NFHOME`, etc. are set in the environment. The `autoconf` build system, however, does not need it: the flags and the path to NetCDF Fortran are supposed to be provided on the command line when calling the configure script. I now realise that my comments are misleading and I also should try to get rid of `RFMIP_CACHEDIR` to make it consistent.

.github/workflows/module_switcher
skosukhin: I have added a comment in the file about what the shell function does.

skosukhin:
This also adds testing with `ifx`. The compiler flags might require some adjustments though.

.github/workflows/containerized-ci.yml
```diff
@@ -1,84 +1,100 @@
...
     strategy:
       fail-fast: false
       matrix:
+        fortran-compiler: [ifort, ifx, nvfortran]
         rte-kernels: [default, openacc]
-        container: ["earthsystemradiation/rte-rrtmgp-ci:ifort","earthsystemradiation/rte-rrtmgp-ci:nvfortran"]
+        include:
+        - fortran-compiler: ifort
+          fcflags: "-m64 -g  -traceback -heap-arrays -assume realloc_lhs -extend-source 132 -check bounds,uninit,pointers,stack -stand f08"
+          image: "earthsystemradiation/rte-rrtmgp-ci:ifort"
```
RobertPincus, line 21: Thanks, this is very elegant

RobertPincus (approved):

skosukhin:
Several more changes:
1. I have to take https://github.com/earth-system-radiation/rte-rrtmgp/pull/200#discussion_r1064809323 back: it's a bad idea to set variables in the matrix if it's possible not to do that. I have stumbled upon several unwanted side effects, which were not so straightforward to workaround. So, I suggest the following policy for matrix and environment variables: don't use them if possible. For example, if an environment variable is used in one step only, it should not be set. Instead, the value should be used in the step literally. If an environment variable has to be set but all jobs need the same value, it should be set directly, without extra matrix variables.
2. The `Self-hosted CI` workflow is now marked green in the [Actions](https://github.com/earth-system-radiation/rte-rrtmgp/actions) although `cce-gpu-openmp` expectedly fails. The workflow is supposed to be red if any other job fails. Note that commits and merge requests are still marked red.
3. A couple of Python scripts now make several attempts (instead of just one) to download files before failing. It looks like the server rejects connections when there are too many of them, which sometimes happens when multiple CI jobs are started at once.

PR #202, opened 2023-01-06
Aerosol Optics MERRA

mjiacono:
Pull request to merge components for deriving aerosol optical properties based on MERRA aerosol optics data.


RobertPincus:
Hi @mjiacono Thanks for this. Two requests. 

1. Can you please change the target of the pull request to the new `feature-merra-aerosol-optics` branch? I will make some changes (adding CI, putting reference files outside the repo) before merging to `develop`. 
2. Can you please add the ability to automatically run the tests and verify the results? That means adding a `tests` target to your `Makefile` that runs the code and a `compare-to-reference.py` file that exits with an error code if the results produced by the code don't match the input files. 

Thanks - Robert

mjiacono:
1. Target changed to the new branch
2. I'll work on this

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,681 @@
...
+  ! ------------------------------------------------------------------------------
+  !
+  ! Compute single-scattering properties
+  !
+  function aerosol_optics(this, aero_type, aero_size, aero_mass, relhum, &
+                          optical_props) result(error_msg)
+
+    class(ty_aerosol_optics), &
+              intent(in  ) :: this
+    integer,  intent(in  ) :: aero_type(:,:)   ! MERRA2/GOCART aerosol type
```
RobertPincus, line 292: @mjiacono Please add comments to clarify the physical meaning of the dimensions here (I'm guessing col, level). Is only one aerosol type permitted per location?

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,681 @@
...
+              intent(inout) :: optical_props
+                                               ! Dimensions: (ncol,nlay,nbnd)
+
+    character(len=128)      :: error_msg
+    ! ------- Local -------
+    logical(wl), dimension(size(aero_type,1), size(aero_type,2)) :: aeromsk
+    real(wp),    dimension(size(aero_type,1), size(aero_type,2), size(this%aero_dust_ext,2)) :: &
+!    real(wp),    dimension(size(aero_type,1), size(aero_type,2), this%get_nband()) :: &
+                 atau, ataussa, ataussag
+                ! ltau, ltaussa, ltaussag, itau, itaussa, itaussag
```
RobertPincus, line 307: @mjiacono Please remove dead code (commented-out lines)

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,681 @@
...
+    end if
+
+    ncol = size(aero_type,1)
+    nlay = size(aero_type,2)
+    nbin = size(this%aero_bin_lo,1)
+    nrh = size(this%aero_rh,1)
+!    nbnd = this%get_nband()
+    nbnd = size(this%aero_dust_ext,2)
+
+    write(61,*) 'Aerosol Optics: checking arrays...'
```
RobertPincus, line 333: @mjiacono Please remove write/print statements. Classes should not input or output.

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,681 @@
...
+    !$acc      create(liqmsk,icemsk)
+    !$omp target data map(to:clwp, ciwp, reliq, reice) &
+    !$omp map(alloc:ltau, ltaussa, ltaussag, itau, itaussa, itaussag) &
+    !$omp map(alloc:liqmsk, icemsk)
+    !
+    ! Aerosol mask; don't need aerosol optics if there's no aerosol
+    !
+    !$acc parallel loop gang vector default(none) collapse(2)
+    !$omp target teams distribute parallel do simd collapse(2)
+    write(61,*) 'Aerosol Optics: defining aerosol mask...'
```
RobertPincus, line 373: No output

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,681 @@
...
+    do ilay = 1, nlay
```
RobertPincus, line 374: @mjiacono It will likely be more efficient to check this inside the interpolation rather than pre-computing the mask

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,681 @@
...
+    end if 
+    !$acc end data
+    !$omp end target data
+  end function aerosol_optics
+  !--------------------------------------------------------------------------------------------------------------------
+  !
+  ! Inquiry functions
+  !
+  !--------------------------------------------------------------------------------------------------------------------
+  function get_aerosol_size_bin(nbin, aero_bin_lo, aero_bin_hi, aero_size) result(ibin)
```
RobertPincus, line 480: Can you please add comments to say what this code is intended to do?

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,681 @@
...
+    real(wp),                  intent(in   ) :: aero_size
+
+    integer  :: i, ibin
+
+    character(len=128)      :: error_msg
+
+    ibin = 0
+
+    error_msg = ''
+    if (aero_size .lt. aero_bin_lo(1) .or. aero_size .gt. aero_bin_hi(nbin)) then
```
RobertPincus, line 492: @mjiacono Please use `any_vals_outside()` from `mo_rte_array_until`

RobertPincus:
@mjiacono How are users to know what integers map to what aerosol type? Is this documented somewhere? In any case we should provide some "introspection" i.e. a way to use variables names (`MERRA_BLACK_CARBON`) instead of integers. 

If we were free to make our own mapping it could open up some efficiencies, like grouping together all the tables where the properties depend on relative humidity.

RobertPincus:
@mjiacono Thanks for including variable `bnd_limits_wavenumber` in the netcdf files. Still shy from the recent cloud physics update, can you confirm a) that the SW data are in strictly increasing order of wavenumber and b) that the `bnd_limits_wavenumber` accurately reflects the data in the fields?

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: Added array dimensions for aero_type, aero_size, aero_mass, and relhum. All are (ncol,nlay). One aerosol type is allowed per layer, but different types can be selected across layers.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: Unneeded commented code removed.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: All write statements removed.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: $acc/$omp directives updated. Write statement removed.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: Aeromsk is pre-computed here to follow what is done in the cloud optics extension.  No change made at this time.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: Function purpose comment added.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: Size check changed to use any_vals_outside()

mjiacono:
The SW aerosol optics coefficient tables are in the proper increasing wavenumber order for RRTMGP in agreement with the values in bnd_limits_wavenumber.

extensions/aerosol_optics/mo_aerosol_optics.F90
RobertPincus: @mjiacono One thing we've learned since we wrote the cloud optics is that the codes are all memory-bound, meaning that it's more efficient to put a comparison, especially a simple comparison, close to where it's needed, so we don't have to access arrays of size (ncol, nlay) multiple times.

RobertPincus:
> @RobertPincus The aerosol types are defined in the ty_aerosol_optics class. I added comments to identify the aero_type integer values in the aerosol_optics function. We can consider converting aero_types from integers to character names, but wouldn’t it be less efficient (in terms of memory) to store character string variable names over an (ncol,nlay) array rather than integers?

@mjiacono I can see two opportunities here. For clarity, what if you move the private definitions of the mapping between integer and aerosol type to somewhere users can see them? I.e. in the module header, not inside the type definition:
    ! MERRA aerosol type 
    integer, parameter, public :: merra_dust_aero = 1
(I think you should add "merra" to be explicit and avoid name clashes.  

We could be fancier and make some data structure that users can inquire the number of aerosol types and their names, but maybe that's not gilding the lily. 

With this in place we can think about combining some of the 7x3 tables.

RobertPincus:
@mjiacono Note that the "continuous integration in a box" is failing because the software containers are now out of sync with the CI; that's my problem not yours.

mjiacono:
> > @RobertPincus The aerosol types are defined in the ty_aerosol_optics class. I added comments to identify the aero_type integer values in the aerosol_optics function. We can consider converting aero_types from integers to character names, but wouldn’t it be less efficient (in terms of memory) to store character string variable names over an (ncol,nlay) array rather than integers?
> 
> @mjiacono I can see two opportunities here. For clarity, what if you move the private definitions of the mapping between integer and aerosol type to somewhere users can see them? I.e. in the module header, not inside the type definition:
> 
> ```
>     ! MERRA aerosol type 
>     integer, parameter, public :: merra_dust_aero = 1
> ```
> 
> (I think you should add "merra" to be explicit and avoid name clashes.
> 
> We could be fancier and make some data structure that users can inquire the number of aerosol types and their names, but maybe that's not gilding the lily.
> 
> With this in place we can think about combining some of the 7x3 tables.

MERRA aerosol types renamed and declarations moved before the ty_aerosol_optics class within the module header.

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,633 @@
...
+! Use and duplication is permitted under the terms of the
+!    BSD 3-clause license, see http://opensource.org/licenses/BSD-3-Clause
+! -------------------------------------------------------------------------------------------------
+! Provides aerosol optical properties as a function of effective radius for the RRTMGP bands
+!   Based on climatoligical aerosol optical properties used in MERRA2 as derived from
+!     the GOCART model for 15 aerosol types, including five dust types, five sea salt types,
+!     one sulfate type, and both hydrophobic and hydrophilic black carbon and organic carbon.
+!   Input aerosol optical data are stored in look-up tables.
+!
+!   References for gocart interactive aerosols:
```
RobertPincus, line 18: @mjiacono Please provide DOIs

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,633 @@
...
+
+  private
+  ! -----------------------------------------------------------------------------------
+  type, extends(ty_optical_props), public :: ty_aerosol_optics
+    private
+    !
+    ! Lookup table information
+    !
+    ! Table upper and lower size bin limits and effective (center) values
+    real(wp),dimension(:), allocatable :: aero_bin_hi(:)
```
RobertPincus, line 57: @mjiacono The `ty_gas_optics` class of which this class is a child has a [way of describing the spectral discretization](https://github.com/earth-system-radiation/rte-rrtmgp/blob/83ff9b315953506333c9143c0b6538d8f7020244/rte/mo_optical_props.F90#L79). Can we adopt the same conventions, i.e. a single vector describing the bounds of the bins? 

Please follow [RTE conventions](https://earth-system-radiation.github.io/rte-rrtmgp/reference/index.html) and use MKS (meters/kilograms/seconds)

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,633 @@
...
+                                    this%aero_bcar_rh_ext, this%aero_bcar_rh_ssa, this%aero_bcar_rh_asy, &
+                                    this%aero_bcar_ext, this%aero_bcar_ssa, this%aero_bcar_asy, &
+                                    this%aero_ocar_rh_ext, this%aero_ocar_rh_ssa, this%aero_ocar_rh_asy, &
+                                    this%aero_ocar_ext, this%aero_ocar_ssa, this%aero_ocar_asy, &
+                                    atau, ataussa, ataussag)
+
+      endif
+
+      !
+      ! Combine liquid and ice contributions into total cloud optical properties
```
RobertPincus, line 419: @mjiacono Can you revise the comment to make it appropriate for the aerosols?

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,633 @@
...
+        error_msg = "aerosol optics: relhumn has wrong extents"
+      if(optical_props%get_ncol() /= ncol .or. optical_props%get_nlay() /= nlay) &
+        error_msg = "aerosol optics: optical_props have wrong extents"
+      if(error_msg /= "") return
+    end if
+
+    !
+    ! Spectral consistency
+    !
+    if(check_values) then
```
RobertPincus, line 353: @mjiacono It seems that if users provide an integer value of aerosol_type that isn't in their list those entries get ignored. Is that the behavior you want? You could also add a check here to be sure the aerosol_type varies from 0 to the max value of aerosol type.

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,633 @@
...
+          ibin = get_aerosol_size_bin(nbin, aero_bin_lo, aero_bin_hi, size(icol,ilay))
+          if (mask(icol,ilay) .and. rh(icol,ilay) .le. rh_offsets(2)) then
+            index = min(floor((rh(icol,ilay) - rh_offsets(1))/rh_step_sizes(1))+1, rh_nsteps(1)-1)
+            fint = (rh(icol,ilay) - rh_offsets(1))/rh_step_sizes(1) - (index-1)
+          elseif (mask(icol,ilay) .and. rh(icol,ilay) .gt. rh_offsets(2)) then
+            index = min(floor((rh(icol,ilay) - rh_offsets(2))/rh_step_sizes(2))+1, rh_nsteps(2)-1)
+            fint = (rh(icol,ilay) - rh_offsets(2))/rh_step_sizes(2) - (index-1)
+          endif
+          ! dust
+          if (mask(icol,ilay) .and. type(icol,ilay) == merra_aero_dust) then
```
RobertPincus, line 556: @mjiacono SELECT CASE is a more transparent way to select from these 7 choices

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,633 @@
...
+            t   = mass(icol,ilay) * tau_dust_table(ibin,ibnd)
```
RobertPincus, line 557: @mjiacono There's a lot of repeated code here. Why not make two interpolation routines, one that depends on RH and the other only on size, then pass table and index to each?

RobertPincus: And following up, why interpolation to tau, tau*ssa, tau*ssa*g, and then normalize in a separate step? Since only one aerosol type is allowed per cell, why not compute the desired output directly and write it into the output arrays right away?

RobertPincus:
@mjiacono There are three tables per aerosol type, one each for extinction, single-scattering albedo, and asymmetry parameter. This makes 21 tables total - a lot to keep track of. 

Since each property for a given type depends on band, band and size, or band, size, and RH. Better to have a single table for each aerosol type, where the last index is 1:3. Define (private) integer parameters `ext_index, ssa_index`, and `asy_index`. Replace references to e.g. `tau_salt_table, ssa_salt_table, asy_salt_table` with `salt_table(:, :, :, ext_index),  salt_table(:, :, :,ssa_index), salt_table(:, :, :, asy_index)` Or make the first index correspond to the quantity, return the three interpolated values as a size-3 vector, and unpack those into the output. 

It would also be worth changing the netCDF file to have 7 tables rather than 21.

RobertPincus:
@mjiacono You could go ahead and synchronize your repo with the main RTE-RRTMGP repo with the "sync fork" button just above the list of code. This will make your continuous integration stop failing for the wrong reasons.

mjiacono:
> @mjiacono You could go ahead and synchronize your repo with the main RTE-RRTMGP repo with the "sync fork" button just above the list of code. This will make your continuous integration stop failing for the wrong reasons.

@RobertPincus My fork of the rte-rrtmgp repo has been synchronized with the main repo for several days. The "sync fork" button reports: "This branch is not behind the upstream earth-system-radiation:main"

RobertPincus:
> > @mjiacono You could go ahead and synchronize your repo with the main RTE-RRTMGP repo with the "sync fork" button just above the list of code. This will make your continuous integration stop failing for the wrong reasons.
> 
> @RobertPincus My fork of the rte-rrtmgp repo has been synchronized with the main repo for several days. The "sync fork" button reports: "This branch is not behind the upstream earth-system-radiation:main"

??? `main` in the RTE repo was updated at 11 am this morning. This is not visible?

mjiacono:
> > > @mjiacono You could go ahead and synchronize your repo with the main RTE-RRTMGP repo with the "sync fork" button just above the list of code. This will make your continuous integration stop failing for the wrong reasons.
> > 
> > 
> > @RobertPincus My fork of the rte-rrtmgp repo has been synchronized with the main repo for several days. The "sync fork" button reports: "This branch is not behind the upstream earth-system-radiation:main"
> 
> ??? `main` in the RTE repo was updated at 11 am this morning. This is not visible?

Your change to main became visible to the fork a few minutes ago. It's synced again.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: @RobertPincus Please clarify “DOI”

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: @RobertPiincus Replaced aero_bin_lo(nbin) and aero_bin_hi(nbin) with merra_aero_bin_lims(npair,nbin) and removed the former arrays and aero_bin_eff (this is the effective center of each size bin, which was provided by MERRA, but we won’t need).

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: @RobertPiincus Comment changed.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: @RobertPiincus Check on aerosol type added. Error returned if out of bounds.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: @RobertPincus Changed aerosol type selection to a SELECT(CASE) structure.

extensions/aerosol_optics/mo_aerosol_optics.F90
RobertPincus: DOIs are Document Object Identifiers, easily translatable into a precise citation. For Chin, for example, the DOI is 10.1029/2000JD900384, and you can access the paper as http://doi.org/10.1029/2000JD900384

mjiacono: @RobertPincus Publication references updated to include DOIs.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: @RobertPincus Understood, but aeromsk is needed where it’s created to limit the bounds checking on the incoming aero_size to elements where aerosols are present.  No change made.

RobertPincus: @mjiacono I don't quite understand. The mask is created by looking at the value of aerosol type indicating 0. The assignment of aerosol properties has to do the same thing. If you have a CASE DEFAULT that sets optical properties to 0 this achieves the same effect but only requires looking at the aerosol type once.

mjiacono: @RobertPincus I see where you’re going now, but it throws me when you pull the code in a different direction from the way you (or Dustin) rewrote the cloud_optics extension.  The array aeromsk has been removed from “compute_all_from_table”. However, the bounds check on aero_size has to remain masked.

RobertPincus: @mjiacono Thanks for the change and agreed that it only make sense to check sizes where there are aerosols defined.  I appreciate the desire to follow the structure of the cloud optics; keep in mind that we have lots of experience esp. in performance since then

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: @RobertPincus Interpolation is only done on the rh dimension. For size, the appropriate bin is selected, since there are only five bins of unequal extent. Only four of the seven aerosol types are interpolated on rh:  salt is 3-D (size, rh, band), and sulfate, bcar_rh, and ocar_rh are 2-D (rh, band). I can add 2-D and 3-D functions to interpolate those tables, but it’s not clear if you prefer calling these functions where needed within the existing loops on band, layer, and column, or pre-interpolating (with the mask) before these loops, putting the result in new arrays, and using those interpolated table values to define the intermediate optics.  

Regarding your second comment, you’ll recall that the aerosol extension is agnostic to longwave or shortwave and so has to handle 1-stream (absorption) or 2-stream (extinction) optics, which it currently does in the separate step. Do you propose building that complexity into the interpolation step?

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,639 @@
...
+    ! Allocate LUT parameters
+    allocate(this%merra_aero_bin_lims(npair,nbin))
+    this%merra_aero_bin_lims = merra_aero_bin_lims
+
+    allocate(this%aero_rh(nrh), &
+             this%rh_nsteps(2), &
+             this%rh_step_sizes(2), &
+             this%rh_offsets(2))
+    this%aero_rh = aero_rh
+    this%rh_nsteps(1) = 17
```
RobertPincus, line 183: @mjiacono The relative humidity discretization should be encoded in the data file and ready from there. No data should be hard-coded.

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -0,0 +1,639 @@
...
+    this%rh_nsteps(2) = 20
```
RobertPincus, line 184: This particular encoding - step sizes and offsets - seems more fragile than a simple list of values.

extensions/aerosol_optics/mo_aerosol_optics.F90
RobertPincus: @mjiacono Thanks for the explanations and the good questions. They prompted me to look a bit more carefully at the treatment of RH and make some suggestions. 

Re interpolation: It seems like you are currently computing both the size index and the rh index and weight for every cell that contains any aerosols. That seems like a good strategy. 

My suggestion was that you write an interpolation function like: 
```
function linear_interp(table, index, weight) result(val)
  real(wp), dimension(:), intent(in) :: table 
  integer, intent(in) :: index 
  real(wp), intent(in) :: weight
  real(wp) :: val 
  
  val = table(index) + weight * (table(index+1) - table(index)))
  end function linear_interp
```
And use this in your `compute_all_from_table` function, i.e. 
```
t   = mass(icol,ilay) * interp(tau_salt_table(ibin, :, ibnd))
```
The : notation extracts the slice of the `tau_salt_table` corresponding to the rh dimension at the values of `ibin` and `ibnd`. This would be more efficient if you reordered the tables to have the RH dimension first (then you'd be passing continuous bits of memory to the function). 

Is that clear?

RobertPincus: @mjiacono Re: the number of tables, you could make the first dimension of every table be `quantity` of length 3. Your interpolation routine could then return a vector of length 3 instead of a scalar. 

If you were being asked for absorption optical depth only it's true that you'd only make use of the first two elements. But the cost of interpolating the extra quantities is tiny compared to the gain in simplicity.

mjiacono: @RobertPincus Thanks for clarifying your preferred direction, and yes, I'll reduce the number of tables.

mjiacono:
> @mjiacono There are three tables per aerosol type, one each for extinction, single-scattering albedo, and asymmetry parameter. This makes 21 tables total - a lot to keep track of.
> 
> Since each property for a given type depends on band, band and size, or band, size, and RH. Better to have a single table for each aerosol type, where the last index is 1:3. Define (private) integer parameters `ext_index, ssa_index`, and `asy_index`. Replace references to e.g. `tau_salt_table, ssa_salt_table, asy_salt_table` with `salt_table(:, :, :, ext_index), salt_table(:, :, :,ssa_index), salt_table(:, :, :, asy_index)` Or make the first index correspond to the quantity, return the three interpolated values as a size-3 vector, and unpack those into the output.
> 
> It would also be worth changing the netCDF file to have 7 tables rather than 21.

@Robert Pincus I completed changes to the code and data files to consolidate the aerosol tables.

extensions/aerosol_optics/mo_aerosol_optics.F90
```diff
@@ -124,51 +124,48 @@ function load_lut(this, band_lims_wvn, &
...
-    if(size(aero_dust_ext, 2) /= nband) &
-      error_msg = "aerosol_optics%init(): array aero_dust_ext has the wrong number of bands"
-    if(size(aero_salt_ext, 3) /= nband) &
-      error_msg = "aerosol_optics%init(): array aero_salt_ext has the wrong number of bands"
-    if(size(aero_sulf_ext, 2) /= nband) &
-      error_msg = "aerosol_optics%init(): array aero_sulf_ext has the wrong number of bands"
-    if(size(aero_bcar_ext, 1) /= nband) &
-      error_msg = "aerosol_optics%init(): array aero_bcar_ext has the wrong number of bands"
-    if(size(aero_ocar_ext, 1) /= nband) &
+    if(size(aero_dust_tbl, 1) /= nval) &
```
RobertPincus, line 137: @mjiacono This seems much more clear to me now, hopefully also to you. 

Can you use `check_extents` from `mo_rte_array_util` please?

RobertPincus:
@mjiacono The most recent round of changes look very good, I hope also to you. Seems to me there are a countably small number of changes to be made: 

1. Using `check_extents` instead of repeated calls to `size`
2. Encoding the RH discretization in the file rather than in code. (Revisiting the way the discretization is a good idea but optional.) 
3. [Encapsulating the interpolation ](https://github.com/earth-system-radiation/rte-rrtmgp/pull/202#discussion_r1069681206)  and simplifying a) so as not to produce intermediate values of tau*ssa and tau*ssa*g, and b) to assign the interpolated values directly to the output fields to avoid unneeded copies. I see you've started on this. 

Otherwise things seem pretty close, thanks very much.

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: Done

RobertPincus: Can you make the same changes at line 273?

mjiacono:
> @mjiacono The most recent round of changes look very good, I hope also to you. Seems to me there are a countably small number of changes to be made:
> 
>     1. Using `check_extents` instead of repeated calls to `size`
> 
>     2. Encoding the RH discretization in the file rather than in code. (Revisiting the way the discretization is a good idea but optional.)
> 
>     3. [Encapsulating the interpolation ](https://github.com/earth-system-radiation/rte-rrtmgp/pull/202#discussion_r1069681206)  and simplifying a) so as not to produce intermediate values of tau_ssa and tau_ssa*g, and b) to assign the interpolated values directly to the output fields to avoid unneeded copies. I see you've started on this.
> 
> 
> Otherwise things seem pretty close, thanks very much.

@RobertPincus (1) and (2) and (3) are completed.

RobertPincus:
> @mjiacono The most recent round of changes look very good, I hope also to you. Seems to me there are a countably small number of changes to be made:
> 
> 1. Using `check_extents` instead of repeated calls to `size`
> 2. Encoding the RH discretization in the file rather than in code. (Revisiting the way the discretization is a good idea but optional.)
> 3. [Encapsulating the interpolation ](https://github.com/earth-system-radiation/rte-rrtmgp/pull/202#discussion_r1069681206)  and simplifying a) so as not to produce intermediate values of tau_ssa and tau_ssa*g, and b) to assign the interpolated values directly to the output fields to avoid unneeded copies. I see you've started on this.
> 
> Otherwise things seem pretty close, thanks very much.

@mjiacono 

On point 1, see lines 273 and following. 

On point 3, a)  there's still lots of repeated code in `compute_all_from_table` - can you encapsulate this in a subroutine, as suggested? And b) you're still going from tau, ssa, g to tau, tau*ssa, tau*ssa*g and back again unnecessarily?

extensions/aerosol_optics/mo_aerosol_optics.F90
mjiacono: Done

mjiacono:
> > @mjiacono The most recent round of changes look very good, I hope also to you. Seems to me there are a countably small number of changes to be made:
> > 
> > 1. Using `check_extents` instead of repeated calls to `size`
> > 2. Encoding the RH discretization in the file rather than in code. (Revisiting the way the discretization is a good idea but optional.)
> > 3. [Encapsulating the interpolation ](https://github.com/earth-system-radiation/rte-rrtmgp/pull/202#discussion_r1069681206)  and simplifying a) so as not to produce intermediate values of tau_ssa and tau_ssa*g, and b) to assign the interpolated values directly to the output fields to avoid unneeded copies. I see you've started on this.
> > 
> > Otherwise things seem pretty close, thanks very much.
> 
> @mjiacono
> 
> On point 1, see lines 273 and following.
> 
> On point 3, a) there's still lots of repeated code in `compute_all_from_table` - can you encapsulate this in a subroutine, as suggested? And b) you're still going from tau, ssa, g to tau, tau_ssa, tau_ssa*g and back again unnecessarily?

@RobertPincus 
On point 3, a) of the seven tables, there are three different sizes, so you would need three subroutines to do what you're suggesting. I don't see the advantage in terms of amount of code or clarity; b) it's not clear what you're after here; the intermediate values are used to derive the final optics for either 1-stream or 2-stream cases.

RobertPincus:
> 

@mjiacono Thanks for the explanation, I see the point about needing tau and tau*ssa as products in order to fill in either _1scl or 2str optical properties. For g, however, this shouldn't be necessary, since g isn't used at all in the _1scl optical properties. Perhaps you can interpolate tau, tau*ssa, and g? 

Please also introduce private integer parameters `TAU = 1, SSA = 2, G = 3` and replace the hard-coded 1, 2, 3 in the first indexes of the tables used in `interp_all_from_table` with  these parameters. 

If you aren't going to implement `interp_aero_table_2d` and `interp_aero_table_3d` please remove them (indeed please remove any unused code e.g. at line 133)

mjiacono:
> > 
> 
> @mjiacono Thanks for the explanation, I see the point about needing tau and tau_ssa as products in order to fill in either _1scl or 2str optical properties. For g, however, this shouldn't be necessary, since g isn't used at all in the _1scl optical properties. Perhaps you can interpolate tau, tau_ssa, and g?
> 
> Please also introduce private integer parameters `TAU = 1, SSA = 2, G = 3` and replace the hard-coded 1, 2, 3 in the first indexes of the tables used in `interp_all_from_table` with these parameters.
> 
> If you aren't going to implement `interp_aero_table_2d` and `interp_aero_table_3d` please remove them (indeed please remove any unused code e.g. at line 133)

@RobertPincus I addressed these points and added the linear interpolation function. Unless you have other suggestions I'll move on to compiling and testing.

RobertPincus:
@mjiacono The code itself looks ready to merge, thanks for all your efforts. 

Can you add an automated test? Ideally this would be Makefile targets for `tests` and `check` as in the `examples` and `test` directory. `make` in your `aerosol_optics/test` directory will build your code and test harness; `make  tests` will run the tests; `make check` will compare the values to known good output, which you can put in the repo for now. 

I'll ensure that these tests are run every time a change is made to the code.

RobertPincus:
Note that having new code exercise tests is one of the thing we [ask contributors](https://github.com/earth-system-radiation/rte-rrtmgp/blob/main/Contributing.md) to do.

mjiacono:
> @mjiacono The code itself looks ready to merge, thanks for all your efforts.
> 
> Can you add an automated test? Ideally this would be Makefile targets for `tests` and `check` as in the `examples` and `test` directory. `make` in your `aerosol_optics/test` directory will build your code and test harness; `make tests` will run the tests; `make check` will compare the values to known good output, which you can put in the repo for now.
> 
> I'll ensure that these tests are run every time a change is made to the code.

@RobertPincus I am working on setting up a test for aerosol optics, but I haven't used the newer testing infrastructure before and at the moment it isn't building for me. One problem is that I can't access the auxiliary input files for the RFMIP clear sky test because we can't use ftp anymore. I'll try to download them some other way.

mjiacono:
@RobertPincus I updated my fork to include modifications to the all-sky test to include aerosol_optics. The /testing directory in my fork is now obsolete and can be disregarded. Note that I left the cloud testing in place, which was set up to run in two out of every three profiles. I set up the aerosols to run in every other profile, so the results reflect all combinations: clear-only, clear-cloud, clear-aerosol, and clear-cloud-aerosol. I did not revise any of the documentation to reflect the aerosol extension code or testing.  A plot showing the clear-aerosol and clear-only results are attached here. 
flxupdn_lwsw_diff_aerosol

RobertPincus:
@mjiacono Love your approach to testing the aerosols in all combinations of clouds. Thanks also for adding the figure. Could you please `git rm` your testing directory? I'll move the netcdf file with the updated all-sky results to the continuous integration staging area before merging. 

This seems done to me, thanks very much, I'll try to get this merged next week.

mjiacono:
> @mjiacono Love your approach to testing the aerosols in all combinations of clouds. Thanks also for adding the figure. Could you please `git rm` your testing directory? I'll move the netcdf file with the updated all-sky results to the continuous integration staging area before merging.
> 
> This seems done to me, thanks very much, I'll try to get this merged next week.

@RobertPincus Glad you're satisfied with the results. I removed the testing directory. Let me know if you spot anything else.

RobertPincus (approved):
Thanks for all this work.

PR #207, opened 2023-01-12
Update CI: fix linking for 'cce-gpu-openmp'

skosukhin:
1. Make the names of the self-hosted configurations consistent. Once this is merged, we can update the branch protection rules. If the list of tests below looks like I expect it to, it requires configurations we don't have anymore.
2. Fix linking for the `cce-gpu-openmp`. The job now fails when running the executables due to some failed sanity checks in the library's code. However interested in OpenMP can now investigate the issue. First, I would check whether the input values are correctly read from NetCDF files.


RobertPincus (approved):
Thanks for keeping the CI so very clear.

PR #219, opened 2023-06-16
extend workaround for Intel optimization bug to v19 compiler

brian-eaton:
I have run into the same problem addressed in PR#170 on NCAR's cheyenne using the intel compiler with CESM.  The compiler version is `ifort (IFORT) 19.1.1.217 20200306.` which sets the CPP macro `__INTEL_COMPILER` to `1910`, and thus doesn't satisfy the condition currently specified, `__INTEL_COMPILER >= 2021`.  Decreasing the compiler version in the conditional to `1910` allows us to run at our default optimization of `O2`.  Running the unit test suit on cheyenne confirms both the failure and that the suggested patch works for us.


RobertPincus (approved):
Thanks, I hope that helps others...

Issue #156, opened 2022-02-02
RRTMGP and Single Precision


mathomp4:
TL;DR: Is there an "official" way to run RRTMGP in single-precision?

---

So this is a bit of a long ramble of a question so bear with me...

In trying to get RRTMGP v1.5 working in GEOS, we found it was crashing with a segfault...and the previous version didn't seem to the last we tested. 

Eventually, I was able to figure out that, for some reason, the set of Intel Fortran flags we switched to recently--which enable vectorization model-wide--lead to a segfault in RRTMGP, and we made that switch in between tests. Now, I don't think the actual issue is in RRTMGP since changing the flags back to our old non-vectorized flags lets it work. I've tried to track down which of our flags cause the issue, but no luck so far.

But I got curious and thought, "I wonder if I can run RRTMGP in single-precision?" for some extra speed (since part of my job is code optimization). So I changed `mo_rte_kind.F90`, recompiled, and boom. Big crash! But, a bit of GitHub searching led me to [this commit](https://github.com/peterukk/rte-rrtmgp-nn/commit/abba4a9a485041babfb732bec954d62acca9dde0) on the rte-rrtgmp-nn repo of @peterukk. So, I added a `k_min`, made similar edits, and I can run RRTMGP in single precision! Even weirder, it works with *both* non-vectorized and vectorized flags (which *really* makes me think someone else is messing with RRTMGP's memory state.)

And this got me thinking: is there a "correct" way to run single-precision RRTMGP? My naive attempt at just changing `wp` failed, but with the tweak from @peterukk it did seem to work and the output wasn't a different planet...

Chiil:
At the moment it is unsupported, but @peterukk and we got it working in SP in the C++ code. There is some pretty recent discussion here, @peterukk has a few fixes that could be of interest for your: https://github.com/earth-system-radiation/rte-rrtmgp/pull/39

mathomp4:
> At the moment it is unsupported, but @peterukk has some fixes for the Fortran code, and we got it working in SP in the C++ code. There is some pretty recent discussion here, @peterukk has a few fixes that could be of interest for you: #39

Well I feel dumb. Comments from yesterday about this! 😄 

Probably I never saw it because I don't know the right words to search for. I kept looking for `REAL32` or the like. (Can you tell I'm not an atmospheric scientist?)

peterukk:
[...]

PR #222, opened 2023-07-12
Feature-numerics

RobertPincus:
This is mostly a first implementation of single precision (finally addressing #156)  following #39. The stand-alone tests are heavily revised - solutions to equivalent problems (e.g reversed vertical orientation) are tested with respect to numerical precision. Single precision fluxes differ from double precision in sample problems by as much as .13 W/m2.


RobertPincus:
Note that @MennoVeerman's [comment](Menno’s concern? https://github.com/earth-system-radiation/rte-rrtmgp/pull/39#issuecomment-1044254604) is not addressed

charleskawczynski:
Hi @RobertPincus, this looks great! I'm curious what sort of solution tolerance you're achieving with single precision. We're testing RRTMGP.jl with Float32, but the errors are quite large at the moment. cc @sriharshakandala

RobertPincus:
@charleskawczynski Feel free to review the PR :-). 

In the test problems in our suite we see difference in single precision with respect to double precision of up to .13 W/m2. @peterukk has a different formulation for the minimum value of k that reduces errors further. I may implement that and/or other alternatives now that automated testing of single precision is in place.

Chiil (approved):
I have tested the code, and it produces good results. The implemented fixes make tests pass that fail with the `develop` code base.

PR #224, opened 2023-08-04
Feature cloud optics gpoint

mjiacono:
This PR includes code revisions to alternately input and use cloud optics input data by g-point. Example test allsky was revised to demonstrate the new capability.


RobertPincus:
Just to note in public that we will move forward with the PR (and the corresponding data PR) when we have validation information i.e. plots that show the new cloud optics improves cloud radiative effect relative to a high spectral resolution calculation

PR #226, opened 2023-08-21
Cray compiler supports nowadays no_create

jussienko:
Additional workarounds are still needed for getting the code to run with Cray compiler. However, this actually fixes issues in the code, so it might be nice to get this into the main code.

### DIFF
--- a/rte-kernels/accel/mo_rte_solver_kernels.F90
+++ b/rte-kernels/accel/mo_rte_solver_kernels.F90
@@ -151,11 +151,7 @@ contains
     !$acc        data copyin(sfc_srcJac) create(   gpt_Jac) if(do_Jacobians)
     !$omp target data map(to:sfc_srcJac) map(alloc:gpt_Jac) if(do_Jacobians)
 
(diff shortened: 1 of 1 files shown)


RobertPincus:
@vectorflux @fomics These fixes were originally yours, I believe. Any objections to the changes?

vectorflux (changes requested):
These #ifdefs were in for a good reason: CCE does not currently support NO_CREATE. Thus this code will not (yet) work on LUMI, and that is why they were put in in the first place. Do you have access to a HPE-CCE compiler which supports NO_CREATE?  I appears there is no runner for CCE-GPU-OpenACC.

jussienko:
Since CCE 15.0.0 no_create is supported. The default CCE in LUMI is 15.0.0, and 15.0.1 is also available via normal module system.

The current `#ifdef`s do not actually even work as all the variables are not necessarily present (I guess that's why no_create's are used in the first place :-) ).

vectorflux:
Good. We have the go ahead from Jussi, so these changes are fine.

PR #228, opened 2023-09-19
Feature CUDA compatability

RobertPincus:
This PR introduces interfaces without implementations for the computational kernels. This allows the use of kernels in other languages that follow the C interfaces. We test this with the CUDA kernels developed by @Chiil and his collaborators in the Netherlands.


RobertPincus:
@Chiil I wonder what you think about merging this to either `develop` or `main` when the CUDA kernels don't pass all the tests. I could make the CUDA kernel tests as experimental?

RobertPincus:
Closing until working

PR #170, opened 2022-05-20
Introduce a workaround for Intel Fortran Compiler Classic 2021 and later

skosukhin:
This introduces a workaround for the vectorization problem reported in #159.

PR #229, opened 2023-09-29
Fix a typo in an OpenMP directive

skosukhin:
This fixes a typo in the OpenMP directive that I mentioned in https://github.com/earth-system-radiation/rte-rrtmgp/pull/170#issuecomment-1652577997.


RobertPincus (approved):
Thanks for the cleanup.

PR #230, opened 2023-09-29
Workaround self-hosted CI

skosukhin:
Somehow, #227 broke the self-hosted CI. It now fails with:
```console
/users/rpincus/actions-runner-101/externals/node20/bin/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /users/rpincus/actions-runner-101/externals/node20/bin/node)
/users/rpincus/actions-runner-101/externals/node20/bin/node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /users/rpincus/actions-runner-101/externals/node20/bin/node)
```

I can't (and don't know yet how to) fix the issue on the server side, so I tried to downgrade `actions/checkout` back to version 3 in `self-hosted-ci.yml` and it worked. Maybe we should have the workaround in the `develop` for the time being.

I also took the liberty to clean up the CI configuration files a bit.

### DIFF
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -16,17 +16,17 @@ jobs:
       matrix:
         fortran-compiler: [ifort, ifx, nvfortran]
         rte-kernels: [default, accel]
-        fpmodel:     [DP, SP]
+        fpmodel: [DP, SP]
         include:
+        # The tests are not experimental by default:
+        - experimental: false
         # Set flags for Intel Fortran Compiler Classic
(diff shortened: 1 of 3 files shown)


skosukhin:
As soon as earth-system-radiation/containers#15 and this are merged, CI should become green again.

RobertPincus (approved):
Thanks, looks good.

RobertPincus (approved):
Looks good - I had made the same changes in another branch but we'll go ahead and merge.

PR #231, opened 2023-09-29
Fix argument parsing in rfmip-clear-sky drivers

skosukhin:
I think there is a bug in `rrtmgp_rfmip_sw.F90` and `rrtmgp_rfmip_lw.F90`: the `rfmip_file` variable is used before it is updated with the argument from the command line. This is my attempt to fix the problem. I would also like to specify `rfmip_file` on the command line in the `autoconf` branch and 132 characters is not enough for that given the long basename of the file. Therefore, I set the maximum length of the variable to 256.

Ideally, of course, we wouldn't have to have the limit at all since even 256 might not be enough in certain scenarios.


RobertPincus (approved):
Nice catch, thanks a lot.

PR #235, opened 2023-10-23
Accumulated changes including reorganization to main

RobertPincus:
Though we expect several big changes in the coming months - CUDA kernels and some updates to numerics - the changes already on develop seem worth propagating to main and issuing a release over.


RobertPincus:
@skosukhin I'd like to merge this PR by about mid-November. Will your next set of patches to `develop` be available by then?

skosukhin:
@RobertPincus sorry for such a long delay. I think that #234 and #237 are the last pieces that we would like to have in the coming release. Thank you.

PR #237, opened 2023-11-20
Fix minmaxloc workaround

skosukhin:
This fixes the condition for the `minmaxloc` workaround:
1. Cray compiler supports `minloc` and `maxloc` for OpenACC and OpenMP starting version 15.
2. Nvidia compiler supports `minloc` and `maxloc` for OpenACC but not for OpenMP (see #113).


RobertPincus (approved):
Thanks for this careful fix.

PR #247, opened 2023-12-02
Simplify build flags

RobertPincus:
Getting rid of `NCHOME` and `NFHOME`

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -25,7 +25,7 @@ jobs:
       FC: ${{ matrix.fortran-compiler }}
       FCFLAGS: "-ffree-line-length-none -m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan -g -DRTE_USE_CBOOL -DRTE_USE_${{ matrix.fpmodel }}"
       # Make variables:
-      NFHOME: /usr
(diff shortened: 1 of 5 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       FC: ${{ matrix.fortran-compiler }}
       FCFLAGS: "-ffree-line-length-none -m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan -g -DRTE_USE_CBOOL -DRTE_USE_${{ matrix.fpmodel }}"
       # Make variables:
-      NFHOME: /usr
+      FCINCLUDE: -I/usr/include
```
skosukhin, line 28: You might want to remove [this line](https://github.com/earth-system-radiation/rte-rrtmgp/blob/fee55ca842ba3e0d96f4acee86a4036197bff817/.github/workflows/containerized-ci.yml#L55), which became redundant after https://github.com/earth-system-radiation/containers/commit/89d5ed43a430d5884844d05d23806389cc114675 (it's interesting that `gfortran` does not look for the Fortran modules in the default directories but `ifort`, `ifx` and `nvfortran` seem to do so).

RobertPincus: It's a good idea - can you make this change in your next PR for the Lumi CI?

PR #250, opened 2023-12-12
Simplify LW source functions

RobertPincus:
The original formulation of RRTMGP uses separate spectral maps for upwelling and downwelling radiation at each level. Here they are combined as sqrt(map1 * map2). This reduces memory traffic between gas optics and solver by about 25% and makes the solvers more generic. 

Differences introduced to the test cases are of order 0.06 W/m2. This exceeds the threshold we've been using to check for equivalence (7e-4). In this iteration the threshold has been relaxed to allow the tests to pass - is this the right approach, or would it be better to establish new test data?

### DIFF
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -59,7 +59,7 @@ jobs:
       RUN_CMD:
       # https://github.com/earth-system-radiation/rte-rrtmgp/issues/194
       OMP_TARGET_OFFLOAD: DISABLED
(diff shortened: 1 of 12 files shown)


RobertPincus:
@alexeedm I'm especially interested in any thoughts you have on the GPU implementation

RobertPincus:
Flux-errors
Net-and-max-flux-error

Long after the fact, here are comparisons of accuracy for present-day RFMIP atmospheres

PR #256, opened 2024-01-23
CI should fail when individual tests fail

RobertPincus:
Small changes to scripts and in-test tolerances.


alexeedm (approved):
Looks good to me.
Tolerance might be a bit tight, but I don't know a robust way to set it. We may need to check it again when NVHPC 24.1 is released.

skosukhin:
This now works with `NVHPC` versions `22.5`, `23.7` and `23.9` (`-g -O2 -Mrecursive -Mallocatable=03 -Mstack_arrays -Minfo=accel,inline -acc=gpu,verystrict -gpu=cc80`) and `Cray 16.0.1` (`-hacc -hadd_paren -Ktrap=divz,ovf,inv -hflex_mp=intolerant -hfp1 -g`).

skosukhin (approved):

PR #258, opened 2024-01-26
Adjust tolerances, fix error in accelerator source function kernel

RobertPincus:
In the course of adjusting tolerances to match @skosukhin's values I uncovered a bug in the LW source function. This PR includes the bug fix and the adjusted tolerances, so more tests pass (nvfortran 23.9 is still failing with a missing pointer)

### DIFF
--- a/examples/all-sky/all_tests.sh
+++ b/examples/all-sky/all_tests.sh
@@ -1,3 +1,4 @@
+set -eux
 ./rrtmgp_allsky 24 72 1 rrtmgp-allsky-lw.nc \
 	    ${RRTMGP_DATA}/rrtmgp-gas-lw-g256.nc ${RRTMGP_DATA}/rrtmgp-clouds-lw.nc ${RRTMGP_DATA}/rrtmgp-aerosols-merra-lw.nc 
 ./rrtmgp_allsky 24 72 1 rrtmgp-allsky-sw.nc \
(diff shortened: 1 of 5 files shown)


RobertPincus:
Also in this PR is changing the testing and example scripts to fail when individual tests fail (as in #256)

skosukhin:
It looks like this makes `Cray 16.0.1` happy as well.

Regarding the failing `NVHPC 23.9` tests, the following quick and dirty workaround fixes the issue:
```patch
--- a/rrtmgp-frontend/mo_gas_optics_rrtmgp.F90
+++ b/rrtmgp-frontend/mo_gas_optics_rrtmgp.F90
@@ -306,12 +306,20 @@ contains
     ! Interpolate source function
     ! present status of optional argument is passed to source()
     !
+    if(present(tlev)) then
     error_msg = source(this,                               &
                        ncol, nlay, nband, ngpt,            &
                        play, plev, tlay, tsfc,             &
                        jtemp, jpress, jeta, tropo, fmajor, &
                        sources,                            &
                        tlev)
+    else
+    error_msg = source(this,                               &
+                       ncol, nlay, nband, ngpt,            &
+                       play, plev, tlay, tsfc,             &
+                       jtemp, jpress, jeta, tropo, fmajor, &
+                       sources)
+    end if
     if(present(tlev)) then
       !$acc        exit data      delete(tlev)
       !$omp target exit data map(release:tlev)
```

@alexeedm looks like a compiler bug after all

RobertPincus:
@skosukhin Ack! The failing tests made me think the change in 5ac4d0a was safe, but it was not I guess...

PR #259, opened 2024-01-26
Extend Levante CI with NAG

skosukhin:
This adds CI tests with the [NAG compiler](https://nag.com/fortran-compiler/) on [Levante](https://docs.dkrz.de/doc/levante/index.html).

The compiler version and flags are what we currently use to build [ICON](https://www.icon-model.org/) by default.

Two of the four added tests currently fail and, therefore, are marked as `experimental`.

I also think that four is too many and we should keep only two. For example, I would keep `nag-cpu-default-SP` (single-precision, default kernels) and `nag-cpu-accel-DP` (double-precision, accel kernels).

### DIFF
--- a/.github/workflows/gitlab-ci.yml
+++ b/.github/workflows/gitlab-ci.yml
@@ -29,7 +29,7 @@ jobs:
     # Check out GitHub repository
     #
     - name: Check out GitHub repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         fetch-depth: 0
     #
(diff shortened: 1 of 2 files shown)


RobertPincus:
@skosukhin Thanks, this is great. I agree on the limiting of the test suite. For me the main advantage of NAG is that it's quite picky about compilation questions (it nags me). It's more valuable to have it cover both accel and default kernels than worry about DP/SP.

RobertPincus:
I'd also like to understand the runtime error produced in the accel kernels.

RobertPincus (approved):
Very helpful, thanks a lot.

PR #261, opened 2024-01-30
Fix the intent in lw_transport_noscat_up

skosukhin:
This should make the `nag-cpu-accel-DP` CI job pass (as soon as Levante is back online).

The wrong intent leads to the multiplication by `NaN` [here](https://github.com/earth-system-radiation/rte-rrtmgp/blob/1949a8aa9897cc4367dad8df90bc59f92e5d80d9/rte-kernels/accel/mo_rte_solver_kernels.F90#L825).

### DIFF
--- a/.github/workflows/gitlab-ci.yml
+++ b/.github/workflows/gitlab-ci.yml
@@ -73,10 +73,6 @@ jobs:
         include:
           # The tests are not experimental by default:
           - experimental: false
-          - config-name: nag-cpu-accel-DP
-            experimental: true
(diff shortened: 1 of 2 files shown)


RobertPincus:
@skosukhin Got it! Because the boundary condition was applied to `flux_up` the argument has to be `inout`. Excellent catch, thanks to you and to NAG. 

Do you mind to also make the change to the CPU kernel as part of this PR?

RobertPincus:
Scratch that request, @skosukhin, as you point out the intent is already correct in the CPU kernel.

PR #263, opened 2024-02-02
Add initial unit tests (solvers)

RobertPincus:
This PR adds some unit tests for the solvers alone (in `tests/rte[sl]w_unit_tests.F90`). Longwave solvers are tested for correctness in radiative equilibrium and for invariance to choices (vertical orientation, problem size...). Shortwave solvers are tested only for invariance.


RobertPincus:
Note that the standalone solver tests fail for `nvfortran` 24.1 and the default kernels. This is surprising because the same compiler/kernel combination passes similar tests in the end-to-end problem (`tests/check_equivalence`). @alexeedm has promised to take a look in the coming days.

RobertPincus:
@skosukhin What would you think about merging this onto `develop` despite the failing CI for nvidia+default-kernels? I'll chase that down after the fact.

RobertPincus:
Of course @alexeedm If you have any bandwidth to look at the failing `nvfortran 24.1`/default kernels failures I'm grateful. You had thought there might be a compiler bug.

tests/mo_testing_utils.F90
```diff
@@ -0,0 +1,276 @@
...
+    real(wp), optional,       intent(in) :: tol 
+    
+    real(wp) :: tolerance 
+    if (present(tol)) then 
+      tolerance = tol 
+    else
+      tolerance = 2._wp
+    end if 
+
+    allclose_2= all(abs(array1-array2) <= tolerance * spacing(array1))
```
skosukhin, line 69: I think that `spacing` should be calculated based on the reference data, which is often (I don't know if that's always the case) passed as the second argument.

skosukhin (changes requested):
@alexeedm please, take a look at the suggested workaround for NVHPC. It looks like the `-Mbounds` introduces a bug into the generated code.

skosukhin (approved):

tests/mo_testing_utils.F90
RobertPincus: Made a more sweeping change in b0ca961 to clarify reference and test values and to use spacing from the ref values

PR #264, opened 2024-02-09
Reduce use of CI via manual runs

RobertPincus:
This lays the groundwork for reducing the number of times the CI is run. The CI will only happen on PRs but can be invoked manually. 

@skosukhin Do we want to change the `on:` rules?


skosukhin:
Something like this, maybe:
```yml
on:
  push:
    branches:
    - main
    - develop
  pull_request:
    branches-ignore:
    - documentation
  workflow_dispatch:
```

skosukhin (approved):
I think it should work.

PR #265, opened 2024-02-16
Run CI jobs on Lumi via GitLab

skosukhin:
This implements CI jobs that run the tests on the AMD GPU partition of [Lumi](https://www.lumi-supercomputer.eu).

The compiler version and flags are what we currently use to build [ICON](https://www.icon-model.org) by default.

For security reasons, the jobs are skipped for PRs from the forks.


RobertPincus (approved):
Seems great, thanks so much.

PR #266, opened 2024-02-16
Small refactoring for the build system

skosukhin:
- fixes the parallel building in the `tests` directory;
- switches to the [recommended](https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html) way of calling `make` recursively;
- gives the users control over the maximum number of `make` processes (generally, `-j` without a value gives worse performance than with the value set to the number of virtual cores on a machine; it's also not very nice towards other users to spawn the unlimited number of processes on a login node);
- makes the `check` recipes honour the `PYTHON` environment variable when locating the Python interpreter (the `python` executable is not generally available and the users might want to set it either to `python3` or to the path to the executable in a virtual environment).

Depends on #265

### DIFF
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -80,8 +80,8 @@ jobs:
       if: matrix.fortran-compiler != 'ifx' || matrix.rte-kernels != 'accel'
       run: |
         $FC --version
-        make libs
-        make -C build separate-libs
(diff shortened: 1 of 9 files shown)


RobertPincus:
I merged #265. Let me know when this PR is ready, @skosukhin

skosukhin:
@RobertPincus this is ready.

PR #267, opened 2024-02-19
Get rid of red crosses in the CI reports

skosukhin:
This should be the first PR in a long time without the annoying red crosses in the CI report.

Several jobs in our CI workflows are known to fail. Currently, we set `continue-on-error: true` for them. This makes at least the workflow status green. However, the red crosses pop up in the PR and commit CI reports. Getting used to them does not look like a good idea to me. One way of dealing with it is to simply remove the jobs. That would be fine with me but I accept the arguments for keeping them.

Job `cce-gpu-openmp` is known to fail at the runtime. This PR changes it so that it is marked green if it does fail at the runtime and is considered as failed otherwise. That is, the job fails the whole `Self-hosted CI` workflow if it does **NOT** fail against the expectations.

The `ifx`+`accel` jobs are known to fail at the build time due to internal compiler errors. This PR changes them so that they are marked green if they fail at the build time with the `Internal compiler error` substring in the build log. In all other cases, the jobs fail the whole `Continuous integration in a box`.

### DIFF
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -77,28 +77,51 @@ jobs:
         repository: earth-system-radiation/rrtmgp-data
         path: rrtmgp-data
     #
-    # Build libraries, examples and tests
+    # Build libraries, examples and tests (expect success)
     #
-    - name: Build libraries, examples and tests
+    - name: Build libraries, examples and tests (expect success)
+      id: build-success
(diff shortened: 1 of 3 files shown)


RobertPincus (approved):
Is there a "wow" emoji? I appreciate that the PR makes clear what our expectations are. Also we know the current `self-hosted CI` will go away when Daint is retired later this year.

Issue #270, opened 2024-03-21
Unable to run RTE-RRTMGP GPU code on NVIDIA A100 GPU


sjsprecious:
I was unable to run the RTE-RRTMGP GPU code on NVIDIA A100 GPU with either `nvhpc/24.1` or `nvhpc/24.3` compiler.

The error message looks like:
```
./check_equivalence test_atmospheres.nc /glade/derecho/scratch/sunjian/rte-rrtmgp/rrtmgp-data/rrtmgp-gas-sw-g224.nc
 gas optics is for the shortwave
   pressure    limits (Pa):    1.005183574463000         109663.3158428461    
   temperature limits (K):    160.0000000000000         355.0000000000000    
   Intialized atmosphere twice
   Default calculation
   Vertical orientation invariance
   Changing TSI fails
   TSI invariance
   halving/doubling fails
   Incrementing with 1scl fails
   Incrementing with 2str fails
   Incrementing with nstr fails
   Incrementing
Warning: ieee_invalid is signaling
Warning: ieee_inexact is signaling
ERROR STOP 1
```
I am using the same compiler flags suggested by https://github.com/earth-system-radiation/rte-rrtmgp/blob/main/.github/workflows/containerized-ci.yml#L40-L42.

RobertPincus:
Closed with 1b75505, thanks @sjsprecious

PR #271, opened 2024-03-21
Fix GPU error on NVIDIA A100 with nvhpc/24.3

sjsprecious:
This PR fixes the runtime error for NVIDIA A100 GPU with nvhpc/24.3 compiler.

Fix #270

PR #273, opened 2024-03-22
Fix GPU runtime error on NVIDIA A100 with nvhpc/24.3

sjsprecious:
This is a reopen of the previous PR (#271 ) that fixes the runtime error for NVIDIA A100 GPU with nvhpc/24.3 compiler. Merge to the `develop` branch instead.

Fix https://github.com/earth-system-radiation/rte-rrtmgp/issues/270


RobertPincus (approved):
Thanks again, @sjsprecious

Issue #215, opened 2023-05-16
RTE shortwave kernel not vectorizing


peterukk:
Hello,

When I tested the current code on an AMD machine I found that `sw_dif_and_source `is not vectorizing with either ifort 2021.6.0 or GNU Fortran 11.3 compilers.

Intel's optimization report says the following:

```
   LOOP BEGIN at ../rte/kernels/mo_rte_solver_kernels.F90(1389,7)
      remark #15344: loop was not vectorized: vector dependence prevents vectorization. First dependence is shown below. Use level 5 report for details
      remark #15346: vector dependence: assumed ANTI dependence between DIR_FLUX_INC(i) (1465:11) and DIR_FLUX_TRANS(i) (1471:11)
   LOOP END
```

(Ignore line numbers, I've added things - but this is for the version of sw_dif_and_source in the main branch).

Timings confirm `sw_dif_and_source` is very slow due to this:

   ```
                               Called  Recurse     Wall      max      min %_of_clea
  clear_sky_total (SW)           1     -       1.946    1.946    1.946   100.00
    gas_optics (SW)            100     -       0.773 8.83e-03 7.65e-03    39.70
    rte_sw                     100     -       1.173    0.017    0.012    60.26
      sw_2stream               100     -       1.170    0.017    0.012    60.12
        sw_dif_and_source    11200     -       1.109 2.79e-03 9.20e-05    56.97
        sw_adding            11200     -       0.043 8.00e-06 3.00e-06     2.19
```

This can be fixed by adding `!$OMP SIMD` before the ncol loop:

```
   LOOP BEGIN at ../rte/kernels/mo_rte_solver_kernels.F90(1389,7)
      remark #15301: OpenMP SIMD LOOP WAS VECTORIZED
      remark #15442: entire loop may be executed in remainder
      remark #15448: unmasked aligned unit stride loads: 1 
      remark #15449: unmasked aligned unit stride stores: 1 
      remark #15450: unmasked unaligned unit stride loads: 6 
      remark #15451: unmasked unaligned unit stride stores: 4 
      remark #15456: masked unaligned unit stride loads: 3 
      remark #15457: masked unaligned unit stride stores: 3 
      remark #15475: --- begin vector cost summary ---
      remark #15476: scalar cost: 470 
      remark #15477: vector cost: 249.250 
      remark #15478: estimated potential speedup: 1.860 
      remark #15482: vectorized math library calls: 2 
      remark #15486: divides: 3 
[...]

PR #275, opened 2024-04-17
Re-vectorize SW two-stream

RobertPincus:
Adopt suggestions by @peterukk in #215 

This approach makes expensive direct beam computations even when solar zenith angle is < 0, but masks those results out in layers where the direct beam does not reach. This will have practical impacts only in columns where the top of the column is in sunlight but the bottom not (or if users aren't removing columns entirely below the horizon).


RobertPincus:
I've used compiler output to ensure that the loop changed in the PR did not vectorize using `gfortran` and `ifort` but does vectorize after this change.

PR #277, opened 2024-04-18
add dummy commit

Try2Code:


skosukhin:
@Try2Code thank you. This worked as expected, i.e. `GitLab CI` jobs were skipped. I will have to find a way to figure out why that was not the case for #276.

PR #278, opened 2024-04-18
Skip Gitlab CI for dependabot

skosukhin:
`GitLab CI` failed for #276 because `dependabot` does not have access to the secrets and we don't want to give it to it. This PR adjusts the rules for `GitLab CI` so that it is skipped for pull requests authored by `dependabot`. Unfortunately, we will learn whether it works only after merging this PR to `develop`.


skosukhin:
For the record, this did not work because the username is `dependabot[bot]` (see 29cc4b3).

PR #279, opened 2024-04-19
Refactor GitLab CI

skosukhin:
This moves common parts of `levante.yml` and `lumi.yml` to `common.yml`, which should make the maintenance easier.


RobertPincus (approved):
Marvelous, very clean

PR #280, opened 2024-04-19
Updated and simplified LW transport

RobertPincus:
Introduces two changes to long wave noscattering radiative transfer calculations. 

1. Changes first-order Gaussian angular quadrature to "Gauss-Jacobi-5" based on results from [Hogan 2023](https://doi.org/10.1002/qj.4598).  
2. Removes Planck function evaluated at layer centers, reducing the number of large arrays used in the LW solver by 1. 

Results for single-angle calculations are comparable to or slightly more accurate than the original formulation (see below). 

Changes kernel API. Reference data for continuous integration is updated and failure thresholds are set to their original, stricter values.


RobertPincus:
#282 introduced the first of these changes. I'll come back to using only level source functions in the future when I'm in better state to compare cost/accuracy tradeoffs.

PR #281, opened 2024-05-09
Facilitate per-band Tsfc Jacobians by including in ty_fluxes

dr0cloud:
For review by @RobertPincus ... Please see @mathomp4 for git questions:
The motivation is to add surface temperature Jacobians by band. The proposed solution was to add the Jacobians to ty_fluxes and use the same reduce infrastructure to make them available per band, per g-point, or however. It also makes the call to rte_lw() a bit cleaner since the Jacobian no longer appears as an optional item in that call.
I am not wedded to this solution .... just for your review and some discussion. 
Thanks, Peter.

### DIFF
--- a/extensions/mo_compute_bc.F90
+++ b/extensions/mo_compute_bc.F90
@@ -39,7 +39,7 @@ module mo_compute_bc
   ! Extend ty_fluxes to report spectrally-resolved downwelling flux at a single layer
   !
   type, extends(ty_fluxes) :: ty_fluxes_1lev
-    real(wp), dimension(:,:), pointer :: gpt_flux_dn => NULL()    ! (ncol, nlev, nband)
+    real(wp), dimension(:,:), pointer :: gpt_flux_dn => NULL()    ! (ncol, nlev, ngpt)
   contains
     procedure :: reduce => reduce_1lev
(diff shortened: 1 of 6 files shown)


RobertPincus:
Adding here some of the off-line conversations with @dr0cloud: 

The application here to be in producing band-wise top-of-atmosphere Jacobians so the TOA radiation between radiation time steps can be computed quickly. 

The disadvantage to the proposed approach is that it requires a new array of size `ncol*nlay*ngpt` rather than accumulating broadband Jacobians on the fly.  The same output could, in principle,  be obtained by doing a one-layer gas optics calculation with increased surface temperature, recomputing the radiation transfer with the original optical properties and the new surface source function, and using a bespoke `ty_fluxes` class  that computed TOA band-by-band fluxes. This is less efficient but probably not prohibitive, and @dr0cloud is going to look into it.

RobertPincus:
Further offline conversations with @dr0cloud ... the question is whether there's widespread appetite for users to be able to compute anything other than broadband Jacobians. If there are other applications from anyone please note them in the comments, otherwise we're likely not to adopt this PR.

RobertPincus:
As we've not heard from anyone else with a use case for general-purpose Jacobians we will close this PR without merging.

PR #287, opened 2024-05-21
Fortran front-end: encapsulate orientation

RobertPincus:
Introduces a `top_at_1` variable encoding the vertical orientation to `ty_optical_props_arry` i.e. optical properties stored as array of values. The value is set by `gas_optics()` but could also be set by hand with `op%set_top_at_1()`. It's undefined by default which could lead to mischief.


RobertPincus:
Introduces a top_at_1 variable encoding the vertical orientation to ty_optical_props_arry i.e. optical properties stored as array of values. The value is set by gas_optics() but could also be set by hand with op%set_top_at_1(). It's undefined by default which could lead to mischief.

Changes the API for ty_optical_props_arry, rte_lw(), rte_sw()

PR #290, opened 2024-05-31
Set default variables in the makefiles

skosukhin:
This sets the default values for several variables that are important for building and testing. The values can still be overridden with the environment variables.

Hopefully, these changes make it easier for the new users to build the libraries and run the tests for the first time.


RobertPincus (approved):
Thanks. I like that this doesn't change anyone's workflow but simplifies where possible.

PR #291, opened 2024-06-02
ci: enable testing on Windows with GCC

gnikit:
- Compilers are installed via the GA fortran-setup (choco in the backend)
- Version of NetCDF via APT has been replaced with conda, where possible
  and packman on Windows
- Consequently, include and link directories need to be setup at runtime
  for variable expansion to occur
- In the Makefiles changes have been made to properly handle variable expansions
  and bash-like regular expressions

## Additional Notes

- GCC 12+ on the Windows runners causes some rather weird behaviour that I was unable to recreate locally with the same compiler version on a Windows 10 VM. I tried a fair bit of time trying to get to the bottom of it but I am not sure what is going on.
- `environment-noplots.yml` is not used anymore so if you have no use for it I could delete it.
- On Windows I install NetCDF via msys because the conda-forge version has been compiled GFortran 5 and hence its ABI is incompatible **and** the the conda-forge package is not available whilst `use-only-tar-bz2` is `true`.
- I believe the Windows cache for Anaconda is not setup correctly.
- It would be great if we could cache the compilers, especially for Windows.
- You could probably test the Intel compilers in the same Workflow.
- I would suggest enabling backtraces and increasing compile/runtime checks with GCC specifically `-fcheck=all -finit-real=snan -finit-integer=-999999 -finit-character=0 -ffpe-trap=invalid,zero,overflow -fbacktrace`. The `-ffpe-trap=...` options cause for a segfault which I am guessing are indicative of a bug.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -9,30 +9,64 @@ on:
     - documentation
   workflow_dispatch:
 
-defaults:
-  run:
-    # Enable Conda environment by using the login shell:
-    shell: bash -leo pipefail {0}
 
 jobs:
   CI:
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        fortran-compiler: [gfortran-10, gfortran-11, gfortran-12]
+        os: [ubuntu-latest, windows-latest]
         fpmodel: [DP, SP]
+        toolchain:
+          - {compiler: gcc, version: 10}
+          - {compiler: gcc, version: 11}
+          - {compiler: gcc, version: 12}
+        include:
+          - os: ubuntu-latest
+            shell: bash -leo pipefail {0}
+            make: make
+          - os: windows-latest
+            shell: "msys2 -leo pipefail {0}"
+            make: mingw32-make.exe
+        exclude:
+          - os: windows-latest
+            toolchain:
+              compiler: gcc
+              version: 12
+
+    defaults:
+      run:
+        shell: ${{ matrix.shell }}
+
     env:
       # Core variables:
-      FC: ${{ matrix.fortran-compiler }}
       FCFLAGS: "-ffree-line-length-none -m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan -g -DRTE_USE_CBOOL -DRTE_USE_${{ matrix.fpmodel }}"
       # Make variables:
-      FCINCLUDE: -I/usr/include
       RRTMGP_ROOT: ${{ github.workspace }}
       RRTMGP_DATA: ${{ github.workspace }}/rrtmgp-data
       RUN_CMD:
       FAILURE_THRESHOLD: 7.e-4
     steps:
+    #
+    # Setup MSYS2 on Windows
+    #
+    - if: runner.os == 'Windows'
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: MINGW64
+        path-type: inherit
+        install: >-
+          mingw-w64-x86_64-make
+          mingw-w64-x86_64-netcdf-fortran
+    #
+    # Setup Fortran compiler
+    #
+    - uses: fortran-lang/setup-fortran@v1.6.1
+      id: setup-fortran
+      with:
+        compiler: ${{ matrix.toolchain.compiler }}
+        version: ${{ matrix.toolchain.version }}
     #
     # Relax failure thresholds for single precision
     #
@@ -54,23 +88,14 @@ jobs:
         path: rrtmgp-data
         ref: v1.8.1
     #
-    # Synchronize the package index
-    #
-    - name: Synchronize the package index
-      run: sudo apt-get update
-    #
-    # Install NetCDF-Fortran (compatible with all compilers)
-    #
-    - name: Install NetCDF-Fortran
-      run: sudo apt-get install libnetcdff-dev
-    #
     # Cache Conda packages
     #
     - name: Cache Conda packages
       uses: actions/cache@v4
       with:
         path: ~/conda_pkgs_dir
-        key: conda-pkgs
+        key:
+          ${{ runner.os }}-conda-${{ hashFiles('environment-ci.yml') }}
     #
     # Set up Conda
     #
@@ -78,26 +103,52 @@ jobs:
       uses: conda-incubator/setup-miniconda@v3
       with:
         miniforge-version: latest
-        activate-environment: rte_rrtmgp_test
-        environment-file: environment-noplots.yml
-        python-version: 3.11
+        activate-environment: rte_rrtmgp_ci
+        environment-file: environment-ci.yml
         auto-activate-base: false
+        use-mamba: true
         # Use the cache properly:
         use-only-tar-bz2: true
+    #
+    # Install NetCDF-Fortran
+    #
+    - name: Install NetCDF-Fortran via Conda
+      if: runner.os == 'Linux'
+      run: conda install -y -c conda-forge netcdf-fortran
+    #
+    # Setup Linux environment variables
+    #
+    - name: Setup env variables - Linux
+      if: runner.os == 'Linux'
+      run: |
+        echo FCINCLUDE=-I ${CONDA_PREFIX}/include >> $GITHUB_ENV
+        echo LDFLAGS=-L ${CONDA_PREFIX}/lib >> $GITHUB_ENV
+        echo LD_LIBRARY_PATH=${CONDA_PREFIX}/lib:${LD_LIBRARY_PATH} >> $GITHUB_ENV
+    #
+    # Setup Windows environment variables
+    #
+    - name: Setup env variables - Windows
+      if: runner.os == 'Windows'
+      run: |
+        echo FCINCLUDE="-I/mingw64/include" >> $GITHUB_ENV
+        echo LDFLAGS="-L/mingw64/lib" >> $GITHUB_ENV
+        echo LD_LIBRARY_PATH="/mingw64/lib:/mingw64/bin:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
     #
     # Build libraries, examples and tests
     #
     - name: Build libraries
       run: |
         $FC --version
-        make -j4 libs
+        ${{ matrix.make }} -j4 libs
     #
     # Run examples and tests
     #
     - name: Build and run examples and tests
-      run: make -j4 tests
+      run: |
+        ${{ matrix.make }} -j4 tests
     #
     # Compare the results
     #
     - name: Compare the results
-      run: make -j4 check
+      run: |
+        ${{ matrix.make }} -j4 check
--- a/environment-ci.yml
+++ b/environment-ci.yml
@@ -1,0 +1,11 @@
+name: rte_rrtmgp_ci
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+  - python=3.11
+  - numpy
+  - netcdf4
+  - xarray
+  - dask
+  - numpy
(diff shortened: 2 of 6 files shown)


RobertPincus:
@gnikit Thanks so much this is wonderful. Do you mind to set the target of the PR to the new branch feature-add-windows-CI that I just made?

RobertPincus:
@skosukhin Alerting you to this which will help tons in building conda feedstocks for the libraries

gnikit:
> @gnikit Thanks so much this is wonderful. Do you mind to set the target of the PR to the new branch feature-add-windows-CI that I just made?

Yes, no problem. I also just noticed that a few debug commits got pushed in this remote. I will drop them.

As for the conda-forge feedstock, let's make an Issue to track progression and have any necessary discussions.

gnikit:
@RobertPincus there are a series of conflicts with the new base branch, is it behind main by any chance?

RobertPincus:
@gnikit Seems like you branched from `main` before commit 41c5fcd? With that commit `develop` and `main` should have been synced, though there are a few mall changes at first.

gnikit:
That's all right. I can rebase/recreate to `earth-system-radiation:feature-add-windows-ci` just to be safe

RobertPincus:
@skosukhin This PR puts Windows and Linux CI in the same file - I'm wondering if that's wisest, or if we should create Windows-specific files?

gnikit:
> @skosukhin This PR puts Windows and Linux CI in the same file - I'm wondering if that's wisest, or if we should create Windows-specific files?

The majority of the code between the Linux and Windows GA workflows is the same, so you would essentially double your maintenance cost were you to split them in two files.

RobertPincus:
Yes, I saw that @gnikit but it also seems like you have to do some shenanigans to adapt the Windows workflow to Linux (e.g. renaming files) and vice-versa (e.g. using chocoltey? to manage compilers). We find that people use the Linux workflow as a guide to set up local installations so I wonder if something that was more Windows-like would be a better example in the long run? (The big caveat is that Gnu make seems to be very not-like-Windows, but maybe we'll move to CMake or similar over time).

RobertPincus:
No strong opinions on my side by @skosukhin has lots of experience with the CI...

RobertPincus:
@skosukhin ?

examples/all-sky/Makefile
```diff
@@ -57,10 +57,10 @@ tests: rrtmgp_allsky
 	$(RUN_CMD) bash all_tests.sh  
 
 check:
-	$${PYTHON-python} ${RRTMGP_ROOT}/examples/compare-to-reference.py --ref_dir ${RRTMGP_DATA}/examples/all-sky/reference --tst_dir ${RRTMGP_ROOT}/examples/all-sky \
+	$${PYTHON-python} "$(RRTMGP_ROOT)/examples/compare-to-reference.py" --ref_dir "$(RRTMGP_DATA)/examples/all-sky/reference" --tst_dir "$(RRTMGP_ROOT)/examples/all-sky" \
```
skosukhin, line 60: As far as I remember, `${var}` and `$(var)` mean the same thing for GNU Make. Are these changes decorative or does `mingw-w64-x86_64-make` choke with the current syntax?

.github/workflows/continuous-integration.yml
```diff
@@ -54,50 +88,67 @@ jobs:
...
+      run: conda install -y -c conda-forge netcdf-fortran
+    #
+    # Setup Linux environment variables
+    #
+    - name: Setup env variables - Linux
+      if: runner.os == 'Linux'
+      run: |
+        echo FCINCLUDE=-I ${CONDA_PREFIX}/include >> $GITHUB_ENV
+        echo LDFLAGS=-L ${CONDA_PREFIX}/lib >> $GITHUB_ENV
+        echo LD_LIBRARY_PATH=${CONDA_PREFIX}/lib:${LD_LIBRARY_PATH} >> $GITHUB_ENV
```
skosukhin, line 126:
```suggestion
        echo "FCINCLUDE=-I${CONDA_PREFIX}/include" >> $GITHUB_ENV
        echo "LDFLAGS=-L${CONDA_PREFIX}/lib" >> $GITHUB_ENV
        echo "LD_LIBRARY_PATH=${CONDA_PREFIX}/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
```

.github/workflows/continuous-integration.yml
```diff
@@ -54,50 +88,67 @@ jobs:
...
+    #
+    # Setup Windows environment variables
+    #
+    - name: Setup env variables - Windows
+      if: runner.os == 'Windows'
+      run: |
+        echo FCINCLUDE="-I/mingw64/include" >> $GITHUB_ENV
+        echo LDFLAGS="-L/mingw64/lib" >> $GITHUB_ENV
+        echo LD_LIBRARY_PATH="/mingw64/lib:/mingw64/bin:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
```
skosukhin, line 135:
```suggestion
        echo "FCINCLUDE=-I/mingw64/include" >> $GITHUB_ENV
        echo "LDFLAGS=-L/mingw64/lib" >> $GITHUB_ENV
        echo "LD_LIBRARY_PATH=/mingw64/lib:/mingw64/bin:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
```
Is the `LD_LIBRARY_PATH` variable needed here? If so, do we need the `/mingw64/bin` entry?

examples/all-sky/all_tests.sh
```diff
@@ -1,4 +1,10 @@
 set -eux
+if [[ $(uname) == *"NT"* ]]; then
+    # Rename files ending with .exe to the same name without the extension
+    for file in *.exe; do
+        mv "$file" "${file%.exe}"
```
skosukhin, line 5: I think that we should not move the files but set the `EXEEXT` variable to `.exe` when on Windows and honour it in the makefiles and scripts:
```makefile
%$(EXEEXT): %.o
	$(FC) $(FCFLAGS) -o $@ $^ $(LDFLAGS) $(LIBS)
```
```bash
./rrtmgp_allsky${EXEEXT} ...
```
and so on.

.github/workflows/continuous-integration.yml
```diff
@@ -54,50 +88,67 @@ jobs:
...
         auto-activate-base: false
+        use-mamba: true
         # Use the cache properly:
         use-only-tar-bz2: true
     #
+    # Install NetCDF-Fortran
+    #
+    - name: Install NetCDF-Fortran via Conda
+      if: runner.os == 'Linux'
+      run: conda install -y -c conda-forge netcdf-fortran
```
skosukhin, line 117: What is the advantage of installing `netcdf-fortran` with Conda (given that it seems to be non-portable and we do that on Linux only)?

skosukhin (changes requested):
At first, having the tests for Linux and Windows in the same file looked like a good idea to me but the difference in the steps that are run on Linux and Windows is quite big to consider moving the Windows tests to a separate file. The CI scripts are often the best and the most consistent sources of information on how to build and run things. I am afraid, that we might lose that. I don't insist though.

I am also not sure if we need to test on Windows at all. Do we really?

RobertPincus:
@skosukhin The idea of Windows testing is in part to prepare for a Conda feedstock for the libraries. @gnikit was responding to a request I made for help on a Discourse since I have no experience with and no access to Windows development environments. 

@gnikit had asked about moving to CMake (as you had in the past); it's starting to seem like that's a really good idea and worth the hassle of having to learn.

gnikit:
I will try and answer your comments @skosukhin in the coming days but my availability is very limited so I would suggest you push directly onto this branch. To setup the win CI.

examples/all-sky/Makefile
gnikit: `$var` and `${var}` are equivalent with the latter being more robust in defining variable boundaries e.g. `echo ${var}foo`.
`$(var)` is used for command execution in your shell. In this case, the usage of `$(RRTMGP_ROOT)` is incorrect and leads to an erroneous expansion of the variable under `msys2` + `mingw-make`.

There are some action runs in my fork where that is demonstrated, (the link to the Action run currently escapes me).

environment-ci.yml
```diff
@@ -0,0 +1,11 @@
+name: rte_rrtmgp_ci
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+  - python=3.11
+  - numpy
```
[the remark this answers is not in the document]
gnikit: > Why is the name `environment-ci.yml` better than `environment-noplots.yml`

Either names are fine. I didn't want to delete your conda config files, which is why I made another environment.

> What is the benefit of the `channels` entry?

It is a good idea to instruct conda from which channels to fetch your packages, e.g. `netcdf` is not available in `default` it's only available in `conda-forge`. This is true for pretty much all your packages. Channels help make your env reproducible, so IMO you should be specifying it everywhere. Docs [link](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/channels.html) to `channels`. 

---

Unrelated to the previous bit, `numpy` has been added because it is needed in `compare-to-reference.py`.

.github/workflows/continuous-integration.yml
gnikit: These are equivalent I believe, just different styling, so happy to change those.

.github/workflows/continuous-integration.yml
gnikit: Same comment as before for the styling, happy to change it to match.

> Is the `LD_LIBRARY_PATH` variable needed here?

Yes, MSYS is not in the runner's PATH, see [docs](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#notes) and we are not setting the `-Wl,-rpath,/bath/to/libs`. We have to set explicitly `LD_LIBRARY_PATH`.

> If so, do we need the `/mingw64/bin` entry?

This is where the compiler's runtime libraries are. I have no idea what compelled the MSYS packagers to place them there, but I am sure there is a reason which I don't know about.

see output from my fork's Action
 https://github.com/gnikit/rte-rrtmgp/actions/runs/9326569389/job/25675276582#step:4:312

> /mingw64/bin/libgfortran-5.dll

examples/all-sky/all_tests.sh
gnikit: yeap, I like this more and it was also my first implementation, but then I realised I didn't know if something else in your pipelines was consuming these binaries, so I played it safe. Happy to make that change.

.github/workflows/continuous-integration.yml
gnikit: It's actually quite portable, see their conda-forge [feedstock](https://anaconda.org/conda-forge/netcdf-fortran). All their installers across OSs and architectures are at v4.6.1. Since conda and conda-forge are great at managing such dependencies I would be inclined to use them as much as possible in this case.
The problem with Windows is that the `gfortran` version is rather old. You don't want to be testing with an old compiler but for packaging purposes it's fine.

Bigger picture: When you are packaging your feedstock one of the steps is to make sure that all your dependencies i.e. `netcdf-fortran` are available in conda-forge. There are ways around this (e.g. building from source), but the APT package manager is not one of them.

gnikit:
> At first, having the tests for Linux and Windows in the same file looked like a good idea to me but the difference in the steps that are run on Linux and Windows is quite big to consider moving the Windows tests to a separate file.

Completely up to you guys. IMO this way is simpler/easier to maintain, since there is a less duplication, but this is *very subjective*. The right answer here is whatever is easier for you to maintain.

> The CI scripts are often the best and the most consistent sources of information on how to build and run things. I am afraid, that we might lose that. I don't insist though.

 Personally, I prefer a well written section in the README or the docs page, since CI workflows can get rather messy, but again completely up to you. As a side note, I found building the lib extremely easy which was great since it made the task a lot easier.

> I am also not sure if we need to test on Windows at all. Do we really?

For me it makes sense to test it if you expect to have/already have users on Windows. It builds up a sense of security and trust to your users.

gnikit:
@RobertPincus @skosukhin I think I finished with my replies (also saw that this got merged in). Always happy to chat more on this, so feel free to ping me if you need something (I might be slow to reply but I'll do my best).

PR #294, opened 2024-08-15
Fixes for CI

skosukhin:
1. Work around the issue with the signature checker on Lumi: push revisions as branches and not as tags.
2. Reduce code duplication in the CI description.
3. Avoid failing the clean-up when removing a GitLab pipeline that has not been created yet.
4. Work around the issue with the old node runtime on Daint (see https://github.com/actions/checkout/issues/1809).


RobertPincus (approved):
Thank you as always.

PR #300, opened 2024-10-30
Code additions for alternate g-point cloud optics

mjiacono:
This PR includes code revisions to input and use alternate g-point cloud optics input data. The all-sky example test case was revised to demonstrate the new capability. 

Longwave and shortwave flux profile comparisons (between band and g-point clouds optics) were completed for the tropical and mid-latitude winter atmospheric profiles using two cloud cases, one an all liquid low cloud and one an all ice mid-level cloud. The graphics below illustrate impacts of up to one percent in the shortwave liquid cloud case, with smaller impacts in the longwave and in the ice cloud case. 

flxupdn_lwsw_diff_cld_bnd_gpt_mlw
flxupdn_lwsw_diff_cld_bnd_gpt_trp

A comparison was also completed between shortwave flux calculations using the RRTMGP band and g-point cloud optics and the LBLRTM shortwave cloudy flux calculations of Lu et al. (2011; https://doi.org/10.1175/JAS-D-10-05001.1). The LBLRTM fluxes from Table 3 of that paper and the comparable results from RRTMGP are shown in the table below. This comparison is not optimized, since small differences in TSI, cloud physical properties, gaseous absorption coefficients, or other parameters may be present. 

LBL_comp_band_g-point_cloud_optics


RobertPincus:
@mjiacono Two requests. 
1. Please open a matching PR that adds the new cloud absorption files to the `rrtmgp-data` repo. 
2. I propose that, rather than changing any existing data files, you add a new file with the cloud properties by g-point. It should have the same format as the original file (the same dataset names) and differ only in having tables defined on g-points rather than on bands. This will keep code changes to a minimum.

RobertPincus:
See also https://github.com/earth-system-radiation/rrtmgp-data/pull/7

PR #310, opened 2024-12-12
Build with CMake

RobertPincus:
Replace the GNU Make build system with CMake based on work by Makepath. Expands CI to include MacOS (13/Intel and latest/ARM M1) and Windows 2022 with gfortran and conda. Updates Ubuntu runners to 24.04. 

Known problem: does not pass tests with window-2022/gfotran-14/double precision; does not currently compile with `nagfor` 17 on Linux (Gitlab integration) because modules are not found. 

I'm unclear if Gitlab integration is running GPU kernels on the GPUs.


RobertPincus:
Closed via #316

Issue #309, opened 2024-12-11
Ice optics uses diameter, not radius 


RobertPincus:
As we address an issue with [mis-labeled cloud optics data](https://github.com/earth-system-radiation/rrtmgp-data/issues/8) in the data repo we should also modify the cloud optics code to be clear that ice particle sizes are to be supplied as diameter, not radius.

RobertPincus:
Closed with #342

PR #311, opened 2024-12-13
Relabel ice sizes as diameter

RobertPincus:
Tables in the RRTMGP data, and code here, suggested that liquid and ice cloud optical properties were tabulated as a function of radius; in fact ice clouds are tabulated vs. diameter. 

Addresses #309


Chiil (approved):
Looks alright to me. Should there be a test added for this? There is no change anywhere that corrects radius to diameter.

RobertPincus:
@Chiil The names of the axes for ice cloud size is changed to `diam`

RobertPincus:
Superseded by #342

PR #312, opened 2024-12-13
Fix gas optics interpolation

tjahns:
This series of fixes addresses a number of bugs and performance problems the original kernel has. While the code is fine with respect to the Fortran standard, it contains a number of places where contemporary CPUs and GPUs can profit from less latency- and optimizer-sensitive idioms.

@skosukhin This works fine for various experiments I did in ICON with the Intel compiler so I'd really like to see this in production, even if that needs additional work on my side. Since I'm not particularly familiar with the rte-rrtmgp external, I'd appreciate any hint on how to refine this PR.

### DIFF
--- a/rrtmgp-kernels/mo_gas_optics_rrtmgp_kernels.F90
+++ b/rrtmgp-kernels/mo_gas_optics_rrtmgp_kernels.F90
@@ -90,6 +90,7 @@ contains
     real(wp) :: press_ref_trop
     real(wp) :: temp_ref_delta_inv
     real(wp) :: press_ref_log_1
+    real(wp) :: press_ref_log_delta_inv
     ! -----------------
     ! local indexes
     integer :: icol, ilay, iflav, igases_1, igases_2, itropo, itemp, jtemp_
@@ -97,6 +98,7 @@ contains
     press_ref_trop = exp(press_ref_trop_log)
     temp_ref_delta_inv = 1.0_wp / temp_ref_delta
     press_ref_log_1 = press_ref_log(1)
+    press_ref_log_delta_inv = 1.0_wp / press_ref_log_delta
     do ilay = 1, nlay
       do icol = 1, ncol
         ! index and factor for temperature interpolation
@@ -105,7 +107,7 @@ contains
         ftemp(icol,ilay) = (tlay(icol,ilay) - temp_ref(jtemp_)) * temp_ref_delta_inv
 
         ! index and factor for pressure interpolation
-        locpress = 1._wp + (log(play(icol,ilay)) - press_ref_log_1) / press_ref_log_delta
+        locpress = 1._wp + (log(play(icol,ilay)) - press_ref_log_1) * press_ref_log_delta_inv
         jpress(icol,ilay) = min(npres-1, max(1, int(locpress)))
         fpress(icol,ilay) = locpress - float(jpress(icol,ilay))
 
--- a/rrtmgp-kernels/mo_gas_optics_rrtmgp_kernels.F90
+++ b/rrtmgp-kernels/mo_gas_optics_rrtmgp_kernels.F90
@@ -149,7 +149,7 @@ contains
             else
               eta = 0.5_wp
             endif
-            loceta = eta * float(neta-1)
+            loceta = eta * real(neta-1, wp)
             jeta(itemp,icol,ilay,iflav) = min(int(loceta)+1, neta-1)
             feta = loceta - aint(loceta)
             ! compute interpolation fractions needed for minor species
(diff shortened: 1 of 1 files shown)


RobertPincus:
@tjahns If you think these changes are generally beneficial could you also port them to `kernels/accel/mo_gas_optics_rrtmgp.F90` ? 

Do you know for sure that the workaround for Intel `ifort` is no longer needed? What version are you testing with?

tjahns:
I tried the versions installed at DKRZ (2022.0.1 and 2023.2.1). but it's hard to make sure that it works in all configurations and versions. I initially looked into this because the work-around failed for some OpenMP setups.
I'll first look into what's going wrong for nvhpc and then investigate applicability to `kernels/accel/mo_gas_optics_rrtmgp.F90`, now.

tjahns:
Is there a recipe to reproduce what happens during the tests exactly? I can't seem to find the all_tests.sh that seems to run extended tests in the CI setup. When I try to test the same setup as (nvfortran, default, SP), by using the following installed packages on Levante
- /sw/spack-levante/nvhpc-24.5-ipi3ad
- /sw/spack-levante/netcdf-fortran-4.6.1-y65bmi
- /sw/spack-levante/mambaforge-22.9.0-2-Linux-x86_64-kptncg
and configure like this
```
 ../../icon-mpim/externals/rte-rrtmgp/configure --disable-silent-rules --enable-gpu=no --enable-single-precision=yes --enable-tests  FC=nvfortran FCFLAGS='-Mallocatable=03 -Mstandard -Mbounds -Mchkptr -Kieee -Mchkstk' --with-netcdf-root=/sw/spack-levante/netcdf-fortran-4.6.1-y65bmi
 ```
 all tests run successfully in a `make -j4 check`.

**Comment by RobertPincus** [comment]
@tjahns Executing `make libs tests check` on `develop` runs all the tests there are to run. If this command executes successfully the tests have passed.

**Comment by tjahns** [comment]
Okay, my bad: I used the autoconf branch that's currently in use in ICON. I'll retry locally with the develop branch.

**Comment by tjahns** [comment]
I was now able to successfully reproduce the error and, after adding some diagnostics, the resulting deviation is about 2 times the expected value:
```
+ ./check_equivalence test_atmospheres.nc ../rrtmgp-data/rrtmgp-gas-sw-g112.nc
 gas optics is for the shortwave
   pressure    limits (Pa):    1.005184        109663.3    
   temperature limits (K):    160.0000        355.0000    
   Intialized atmosphere twice
   Default calculation
   Vertical orientation invariance
   TSI invariance
 max deviation:    2.4414063E-04  tolerance:    1.2207031E-04
   halving/doubling fails
 max deviation:    2.4414063E-04  tolerance:    1.2207031E-04
   Incrementing with 1scl fails
 max deviation:    2.4414063E-04  tolerance:    1.2207031E-04
   Incrementing with 2str fails
 max deviation:    2.4414063E-04  tolerance:    1.2207031E-04
   Incrementing with nstr fails
   Incrementing
Warning: ieee_invalid is signaling
Warning: ieee_underflow is signaling
Warning: ieee_inexact is signaling
ERROR STOP 1
```
My change to get these values printed:
```patch
diff --git a/tests/check_equivalence.F90 b/tests/check_equivalence.F90
index 7ece94a..67529c6 100644
--- a/tests/check_equivalence.F90
+++ b/tests/check_equivalence.F90
@@ -655,7 +655,8 @@ contains
     real(wp), dimension(:,:), intent(in) :: array1, array2
     real(wp), optional,       intent(in) :: tol
 
-    real(wp) :: tolerance 
+    real(wp) :: tolerance, maxdev
+    integer :: position(2)
     if (present(tol)) then
       tolerance = tol
     else
@@ -663,6 +664,13 @@ contains
     end if
 
     allclose = all(abs(array1-array2) <= tolerance * spacing(array1))
+    if (.not. allclose) then
+      position = maxloc(abs(array1-array2))
+      maxdev = abs(array1(position(1), position(2)) &
+           - array2(position(1), position(2)))
+      write (0, *) 'max deviation: ', maxdev, &
+           ' tolerance: ', spacing(array1(position(1), position(2)))
+    end if
   end function allclose
   ! ----------------------------------------------------------------------------
   subroutine report_err(error_msg)
```

tjahns:
The deviation is a result of the 3 replacements of division by multiplication with the inverse. So I guess it's your call if the deviation introduced is worth relaxing the criterion or if you'd rather stay within the limits currently set. When I revert only be0801c6a5d6bac91f71271b0f001220bb5563cb that's already enough to get the tests to pass in my setup.

RobertPincus:
@tjahns Thanks for this careful investigation. It looks like the variation is precisely a factor of 2 larger than we allow for in the `tolerance` arguments at line 41-414, does that seem right? There's nothing magic about those tolerance arguments - perhaps you can propose the smallest possible increases to the tolerances that allows the tests to pass?

RobertPincus (approved):
Thanks very much for these @tjahns

Issue #314, opened 2024-12-19
Revisit mo_rte_kind.F90


skosukhin:
I think we should make a couple of decisions on [`mo_rte_kind.F90`](https://github.com/earth-system-radiation/rte-rrtmgp/blob/c2a6e438d8b45ff23f24a098e34bb7440ba8e49e/rte-frontend/mo_rte_kind.F90):
1. Drop the `RTE_USE_CBOOL` macro and consistently use the kind from the `iso_c_binding` module like for other types by default. I remember there were compilers (NVHPC?) that had problems with that. If that is true for more recent compilers (say, 23+), we should switch (preferably locally, not globally in `mo_rte_kind.F90`) to the Fortran default logical kind using the compiler-specific `#ifdef`s.
2. We currently have a rather counter-intuitive circular dependency: `rte-kernels` need `mo_rte_kind` residing in `rte-frontend`, which depends on `rte-kernels`. We should probably move `mo_rte_kind.F90` to `rte-kernels`.

RobertPincus:
@skosukhin 
1. If all compilers are happy with using C-compatible  booleans I'm happy to use that, I guess. I can check... but maybe I would wait until the CMake and other PRs are merged? On the other hand if this made the Cmake refactoring easier this could be part of your PR. 
2. I don't mind where `mo_rte_kind.F90` sits; `rte-kernels` is fine.

RobertPincus:
Item 2 is resolved in #316.

RobertPincus:
@skosukhin I had in mind that using the default Fortran boolean kind would be "easier" for compilers than using the C-compatible kind but maybe that's silly, and we should always use the latter if all compilers support it.

RobertPincus:
See #331

PR #307, opened 2024-12-02
Added basic CI for MacOS and Windows

makepath-alex:
The current basic workflow is to use conda
It has been tested on Windows/Linux/Mac and WSL.
Unfortunately still on some systems we encounter errors, which I am trying to figure out as they emerge, but in most cases the workflow works as expected.
Steps to configure and build:

Create conda dev environment:
```
conda env create -f environment-dev.yml
```

Set environment variables:
```
export FC=gfortran
export FCFLAGS="-ffree-line-length-none -m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan"
export RRTMGP_DATA_VERSION=v1.8.2
export FP_MODEL=DP
export RTE_CBOOL=ON
export RTE_KERNELS=default
export FAILURE_THRESHOLD=7.e-4
```

Configure build using Ninja. We use ninja becuase it is multiplatform and thus the setup is identical on Windows, Linux and Mac. Feel free to replace the genertor with another one of your choince, for example `-G "Unix Makefiles"`
```
cmake -S . -B build -G "Ninja" \
        -DCMAKE_Fortran_COMPILER=$FC \
        -DCMAKE_Fortran_FLAGS="$FCFLAGS" \
        -DRRTMGP_DATA_VERSION=$RRTMGP_DATA_VERSION \
        -DPRECISION=$FP_MODEL \
        -DUSE_C_BOOL=$RTE_CBOOL \
        -DKERNEL_MODE=$RTE_KERNELS \
        -DENABLE_TESTS=ON \
        -DFAILURE_THRESHOLD=$FAILURE_THRESHOLD \
        -DCMAKE_BUILD_TYPE=Debug
```
Note that we configure CMAKE for **release** build `-DCMAKE_BUILD_TYPE=Release`. 
Alternatively if you want a debug configuration, change the work Release with Debug and debug flags like `-g` will automatically be added to the compiler. 

The build configuration will be stored in `build` folder and then you can either go there and build the source using your generator's default instruments, or you can build it with Cmake. I prefer telling CMake to build it for me, this way it's the same command line that I can use on all OS and it's agnostic of the generator I use, be it Ninja, Makefiles or MS Visual Studio
```
cmake --build build --target all --parallel
```

Finally, if **ENABLE_TESTS** option was selected, you can also run the tests using
```
ctest --test-dir build/ -V
```

Important note.
Once CMake has been configured, all the build-rated files will be located in the `build` folder.
[...]

PR #316, opened 2024-12-20
Refactor CMake and CI scripting

skosukhin:
This is a refactoring for the CMake scripts (see #305, #307 and #310). Hopefully, it is a bit more idiomatic and predictable (in terms of library location) now.

Abridged summary of changes, side effects and things that need to be done in the future:
1. The root `CMakeLists.txt` now specifies the version of the library. @RobertPincus, please make sure you update [this line](https://github.com/earth-system-radiation/rte-rrtmgp/blob/baa930eeb1f490bd7fb2d40d31525bcc8a6c06af/CMakeLists.txt#L5) whenever you release a new version. Ideally, there should be a separate commit `Bump version to ...` and the respective release tag references it. Please, feel free to add the patch-level version number.
2. I have updated [`setup.sh`](https://github.com/earth-system-radiation/rte-rrtmgp/blob/baa930eeb1f490bd7fb2d40d31525bcc8a6c06af/setup.sh#L12-L18) to demonstrate what command-line arguments are now expected from the user (`FAILURE_THRESHOLD` can be omitted though).
    `TODO:` remove `setup.sh` (unconventional and misleading) and provide the building instructions in `README.md`
3. `NetCDF_Fortran` is now searched for using the `find_package` command, which means that there is a number of environment variables/command-line arguments that can be set/extended to help `cmake` find it (e.g. [`NetCDF_Fortran_ROOT`](https://cmake.org/cmake/help/latest/variable/PackageName_ROOT.html#variable:%3CPackageName%3E_ROOT)).
4. All CI scripts were updated accordingly. The `self-hosted-ci.yml` configuration is removed (it does not work anymore anyway, I guess). A new CI workflow `Check Style` is added. It is supposed to make sure that all CMake scripts are consistently formatted. If the `Format` job fails, it will upload a patch file that one can download and apply to the code to meet the style requirements. The `Lint` job makes sure that things like local variable names are consistent.
    `TODO:` extend the jobs to cover `Python` (`isort`, `black`, `flake8`), `C/C++/CUDA` (`cmake-format`) and `Fortran` (`fprettify`) source files.
5. All known problems mentioned in #310 seem to be resolved now.
6. File `mo_rte_kind.F90` is moved to `rte-kernels` (see #314).
7. `TODO:` come up with a common prefix for custom CMake variables. For example, `BUILD_TESTING` is a standard variable and should keep the name but `KERNEL_MODE`, `FAILURE_THRESHOLD`, `RTE_ENABLE_SP`, etc. are project-specific and should have the same and a project-specific prefix. Usually, this would be the name of the project but `RTE_RRTMGP_` seems to be rather long. Maybe we could use something like `RR_`.
8. The tests are currently missing dependencies on each other and, therefore, we cannot run them in parallel.
    `TODO:` introduce the missing test `FIXTURES` and consider fetching the required data at the build time as one of the tests.
9. `TODO:` make it possible to install the libraries (and maybe the data files).
10. `TODO:` come up with the public aliases for the targets (e.g. `rte-rrtmgp:rte`, `rte-rrtmgp:rte-dp`, etc.).
11. `TODO:` make sure we can build `rte-rrtmgp` as a subproject (i.e. `add_subdirectory(rte-rrtmgp)` should work and expose the targets).
12. `TODO:` make sure that a superproject can find `rte-rrtmgp` (i.e. `find_project(rte-rrtmgp)`) in the build directory, as well as in the installation prefix.
13. `TODO:` consider changing the directory layout. For example, it feels natural to move `rte-kernels` and `rte-frontend` to a new first-level subdirectory `rte`. Directories `gas-optics`, `rrtmgp-kernels` and `rrtmgp-frontend` could be moved to a new first-level subdirectory `rrtmgp`. We also have a minor inconsistency: the kernel mode is named `extern` but the respective files are stored in the `api` subdirectory.
14. `TODO:` revisit `.gitignore`.

There probably will be more `TODO` from me later (we should create an issue for that, I guess).


RobertPincus (approved):
Thanks so much for this @skosukhin

PR #319, opened 2025-01-03
CMake: add missing test dependencies

skosukhin:
This solves several issues related to the tests:
- `examples/all-sky`: currently, CMake copies [files from `${RRTMGP_DATA}`](https://github.com/earth-system-radiation/rte-rrtmgp/blob/e13d20e5b6c3df99188f6564c415976a3438d316/examples/all-sky/CMakeLists.txt#L2-L5) to the build directory at the configure stage. The files are then fully overwritten by `rrtmgp_allsky` at the test stage. So, there is no point in copying. Another problem is that this might lead to false positive results of `compare-to-reference.py`: the script succeeds even if `rrtmgp_allsky` was never run. This PR drops the copying and makes sure that `compare-to-reference.py` runs only when the respective `rrtmgp_allsky` runs are finished successfully (see the `FIXTURES_SETUP`/`FIXTURES_REQUIRED` properties of the tests).
- `examples/rfmip-clear-sky`: we don't need to copy `multiple_input4MIPs_radiation_RFMIP_UColorado-RFMIP-1-2_none.nc` because the test programs don't modify it and accept the path to the file as the command line argument. Unfortunately, this is not the case for all the [other input files](https://github.com/earth-system-radiation/rte-rrtmgp/blob/e13d20e5b6c3df99188f6564c415976a3438d316/examples/rfmip-clear-sky/CMakeLists.txt#L4-L7). Not only it is impossible to specify paths to the files on the command line but the test programs also write to the files. As before, those files are currently copied at the configure stage and there is a lot that can happen before the test stage. Therefore, the PR implements the copying as a separate "test". This way, there is a much higher chance that the input files for the tests are the ones we expect them to be (e.g. if we run `ctest` several times, we now get "fresh" input files each time and not only for the first run).
    `TODO:` consider rewriting the test programs so that they do not write to the input files but produce new ones in the build directory. Ideally, we should also make it possible to specify paths to the files on the command line to fully eliminate the need for copying.
- `tests`: `check_variants` writes to the input file. Therefore, we have to copy `multiple_input4MIPs_radiation_RFMIP_UColorado-RFMIP-1-2_none.nc` to the build directory. Moreover, we have to make sure that `check_variants_lw` and `check_variants_sw` are not run in parallel since they write to the same file (see the `DEPENDS` property of `check_variants_sw`). On the other hand, `check_equivalence` does not need a copy of the file and can run against the original one from `${RRTMGP_DATA}`. Hence, the respective tests don't need to wait for the copying to finish.
    `TODO:` consider rewriting the test programs (and probably `validation-plots.py`) to make the copying redundant.
- as I mentioned above, a lot can happen between the configure and the test stages. Therefore, this PR makes the fetching of the input data from [`rrtmgp-data`](https://github.com/earth-system-radiation/rrtmgp-data) an extra "test", which is required by all tests that need the data. This way, we make sure that the expected tag is cloned from the data repository right before running the tests. It is also nice to be able to configure with `-DBUILD_TESTING=ON` without making the configure stage too long and fetch the data only when running `ctest`. If the users already have the required data locally, they can suppress the fetching by configuring with `-DRRTMGP_DATA=/path/to/local/copy/of/rrtmgp-data`. When the variable is provided, the extra "test" `fetch_rrtmgp_data` becomes a simple `cmake -E true`.
- this PR introduces an internal cache variable `_RTE_RRTMGP_HAVE_PY_PACKAGES` to avoid rerunning the checks for the required Python packages over and over again.
- CI jobs now run the tests in parallel.
- the PR also fixes a couple of things that I forgot to do in #316: the GitLab pipelines now use the same default values for `FAILURE_THRESHOLD` as all other tests and all `gfortran` tests don't need the `-ffree-line-length-none` flag to be set via `FFLAGS` since it is [injected by CMake automatically](https://github.com/earth-system-radiation/rte-rrtmgp/blob/e13d20e5b6c3df99188f6564c415976a3438d316/CMakeLists.txt#L23).

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -61,7 +61,7 @@ jobs:
             fpmodel: SP
     env:
       FC: gfortran-${{ matrix.gfortran-version }}
-      FFLAGS: "-ffree-line-length-none -m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan"
+      FFLAGS: "-m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan"
       CMAKE_BUILD_PARALLEL_LEVEL: 8
       VERBOSE:
       # TODO: add missing test dependencies and run them in parallel:
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -64,8 +64,7 @@ jobs:
       FFLAGS: "-m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan"
       CMAKE_BUILD_PARALLEL_LEVEL: 8
       VERBOSE:
-      # TODO: add missing test dependencies and run them in parallel:
-      # CTEST_PARALLEL_LEVEL: 8
+      CTEST_PARALLEL_LEVEL: 8
       CTEST_OUTPUT_ON_FAILURE: 1
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/containerized-ci.yml
+++ b/.github/workflows/containerized-ci.yml
@@ -56,8 +56,7 @@ jobs:
       NetCDF_Fortran_ROOT: /opt/netcdf-fortran
       CMAKE_BUILD_PARALLEL_LEVEL: 8
       VERBOSE:
-      # TODO: add missing test dependencies and run them in parallel:
-      # CTEST_PARALLEL_LEVEL: 8
+      CTEST_PARALLEL_LEVEL: 8
       CTEST_OUTPUT_ON_FAILURE: 1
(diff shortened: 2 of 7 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -61,11 +61,10 @@ jobs:
             fpmodel: SP
     env:
       FC: gfortran-${{ matrix.gfortran-version }}
-      FFLAGS: "-ffree-line-length-none -m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan"
+      FFLAGS: "-m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan"
```
RobertPincus, line 64: Noting here an exchange I had with @skosukhin, noting that only `gfortran` requires this flag.

RobertPincus (approved):
Thanks as always @skosukhin. This is a big help.

PR #322, opened 2025-01-06
Tests don't write to inputs

RobertPincus:
As per @skosukhin's suggestion in #319, `tests/check_variants` no longer writes to input files but instead creates a new output file. SW and LW versions must still run in sequence.


tests/CMakeLists.txt
```diff
@@ -76,55 +75,35 @@ foreach(g_value IN ITEMS 224 112)
...
+  check_variants_sw PROPERTIES FIXTURES_REQUIRED fetch_rrtmgp_data
 )
 
 add_custom_target(
   validation-plots
   COMMAND
     ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/validation-plots.py
-    --input_file multiple_input4MIPs_radiation_RFMIP_UColorado-RFMIP-1-2_none.nc
+    --state_file
+    ${RRTMGP_DATA}/examples/rfmip-clear-sky/inputs/multiple_input4MIPs_radiation_RFMIP_UColorado-RFMIP-1-2_none.nc
```
skosukhin, line 107: I would specify the `--lw_vars_file` and `--sw_vars_file` arguments explicitly here. It will be easier to understand that the target requires outputs from the tests.

`TODO:` enforce running the test programs when the `validation-plots` target is requested.

skosukhin (approved):
I can't tell much about the Fortran part but CMake LGTM.

tests/CMakeLists.txt
RobertPincus: Good idea, see 9b8d000

Issue #325, opened 2025-01-06
CMake installation includes data files? 


RobertPincus:
@skosukhin proposes in #323 that data files be installed as part of the CMake installation target. 

Moving the discussion from there: in principle we would want all the `*.nc` files in the root of the RRTMGP repository. 

Perhaps it's worth comparing notes with the Python folks e.g. @tcmetzger and @brendancol

tcmetzger:
Let's discuss this during our update call on Wednesday!
CC @sehnem @makepath-alex

PR #326, opened 2025-01-06
CMake: install data files

skosukhin:
This extends the `install` target to install all the `*.nc` files in the root of the [rrtmgp-data](https://github.com/earth-system-radiation/rrtmgp-data) (see #325).

As in the case with the tests, the data files are not fetched at the configure stage but at the latest stage possible.


RobertPincus (approved):
Fine for me.

PR #331, opened 2025-01-07
Always use c_bool for logical kind

skosukhin:
Since `wl` is a logical kind of arguments of interoperable procedures (e.g. [`lw_solver_noscat`](https://github.com/earth-system-radiation/rte-rrtmgp/blob/2c4a848231150d2e2ecddae5c8c418dd090b9a03/rte-kernels/mo_rte_solver_kernels.F90#L377-L381)) it should always be the one from `iso_c_binding` (like all other kinds defined in [`mo_rte_kind.F90`](https://github.com/earth-system-radiation/rte-rrtmgp/blob/2c4a848231150d2e2ecddae5c8c418dd090b9a03/rte-kernels/mo_rte_kind.F90) actually). I think it works for all compilers that we are interested in at the moment. If this becomes an issue at some point, we should tweak the value of `wl` with compiler-version-specific preprocessor guards.

This PR drops the configure time check for whether the compiler in use tolerates non-interoperable logical type of the arguments of interoperable procedures and sets `wl` to `c_bool` unconditionally.


skosukhin:
I will rebase this once #332 is merged.

PR #334, opened 2025-01-08
Revive CI on Lumi

skosukhin:
This enables CI on LUMI with the Cray compiler and AMD GPUs again. The respective GitLab runner is registered at [gitlab.dkrz.de](https://gitlab.dkrz.de/icon-libraries/librte-rrtmgp), and we can have only one mirror now, which significantly simplifies maintenance.


skosukhin:
@RobertPincus I would also like to rename `.gitlab/levante.yml` to `.gitlab/dkrz.yml` but this will break the `GitLab CI` workflow for all branches that do not have these changes. Should I do it now or later?

RobertPincus:
@skosukhin Go ahead and do it now, I think. I can merge the commit onto other branches onto other branches if need be

PR #335, opened 2025-01-08
Optionally produce shared libraries

skosukhin:


skosukhin:
We should probably change some CI jobs to build the shared libraries but I have no idea how to check that we get the expected result.

RobertPincus:
@skosukhin This PR **replaces** the building of static libraries with the building of shared libraries? Any downsides to this?

skosukhin:
When configured with `-DBUILD_SHARED_LIBS=OFF` (by default), the build system produces static libraries as before. If the option is set to `ON`, shared libraries are produced.

RobertPincus (approved):
Sweet!

PR #336, opened 2025-01-13
Support NEC Aurora compiler

skosukhin:
This updates the CMake scripting to support the [`NEC SX-Aurora TSUBASA Compiler`](https://www.nec.com/en/global/solutions/hpc/sx/tools.html) (a.k.a `nfort`).

With `nfort 5.3.0`, all tests pass with `-DCMAKE_BUILD_TYPE=Release -DRTE_ENABLE_SP=OFF` (i.e. `-O3`). With `-DRTE_ENABLE_SP=ON`, all tests pass only with `-O1`. Any higher optimization level makes `check_equivalence_sw_g112` fail with:
```
   Changing TSI fails
   halving/doubling fails
   Incrementing with 1scl fails
   Incrementing with 2str fails
   Incrementing with nstr fails
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,17 +20,24 @@ set_property(CACHE KERNEL_MODE PROPERTY STRINGS ${PREFERRED_KERNEL_MODES})
 
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
+# CMake does not support the NEC compiler and usually mistakes it for GNU:
+if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU)
+  # Check whether the Fortran compiler in use is NEC:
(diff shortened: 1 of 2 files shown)


RobertPincus:
@skosukhin You want to merge the PR even under these circumstances?

skosukhin:
> @skosukhin You want to merge the PR even under these circumstances?

Yes, please. To be honest, I still think that there is something strange with the test. I recall problems with it on LUMI with Cray compiler (magically disappeared in #334). If I remember correctly, it's also sensitive to the version of NVHPC in use (I can't find the exact reason for switching to `NVHPC 24.9` in Levante CI). The test itself might be fine. Maybe the compilers cannot handle it properly. We will see, I guess.

RobertPincus (approved):
@skosukhin Sorry this is so awkward but thanks for the support.

PR #342, opened 2025-01-14
Ice sizes are diameters, not radii

RobertPincus:
Tables in the RRTMGP data, and code here, suggested that liquid and ice cloud optical properties were tabulated as a function of radius; in fact ice clouds are tabulated vs. diameter.

Addresses #309. Originally #311.


RobertPincus:
Merging because these are the same changes as #311 that @Chiil has already approved.

PR #343, opened 2025-01-15
Correct cbind function arguments for Cloud optics kernels

makepath-alex:
Updated the argument order for the following functions to match the Fortran implementation:
- rrtmgp_compute_tau_rayleigh
- rrtmgp_compute_cld_from_table
- rtmgp_compute_cld_from_pade

Removed repeated definition for rrtmgp_compute_tau_rayleigh


RobertPincus (approved):
Ouch, sorry, and thanks for the fix @makepath-alex. 

Note that the Pade kernels will go away in a revision coming soon.

PR #345, opened 2025-01-19
Fix API cloud binds

sehnem:
I was having some issues with these functions in pyRTE-RRTMGP and found that there where some issues with some of the variables, this PR fixes it.


RobertPincus:
@sehnem Can you revise the PR so there are no merge conflicts with the current `develop`?

RobertPincus (approved):
Sorry to have missed the `const` specifications.

PR #347, opened 2025-01-23
Allow for cloud optics defined on g-points

RobertPincus:
Depending on the data file, cloud optics are computed on bands or g-points.


RobertPincus:
Pade approximant cloud optics are removed.

skosukhin (approved):
LGTM with respect to the CMake scripting. However, I don't like the idea of having a [mutable reference](https://github.com/earth-system-radiation/rte-rrtmgp/blob/2089773eea532aebbdf6cfa0ede79d7a710d082d/CMakeLists.txt#L93) (changed in #342) in the script.

RobertPincus:
Nice that I learn an English acronym from you! 

How would you suggest to manage PRs when the data needed to pass the tests changes?

skosukhin:
> How would you suggest to manage PRs when the data needed to pass the tests changes?

The main problem with referencing the `develop@rrtmgp-data` (or any other branch) is that it becomes difficult to work with and run the tests for an older version of `develop@rte-rrtmgp`. Imagine that we are a couple of months in the future. The data in `develop@rrtmgp-data` have changed several more times but are, of course, consistent with `develop@rte-rrtmgp`. All tests pass and everything looks good. However, for whatever reason, I have to work with the `develop@rte-rrtmgp` from today and `develop@rrtmgp-data` from the future does not work for it. I will have to painfully go through the histories of the repositories to figure out the version of the data that fits the version of the library that I'm working on. I have actually been there once (we used to reference the data in the CI description back then though). Rather annoying.

Also, right at the moment, does the current [`develop`](https://github.com/earth-system-radiation/rte-rrtmgp/commit/2089773eea532aebbdf6cfa0ede79d7a710d082d) actually pass the CI if you run it? I expect it to fail because the current `develop@rrtmgp-data` is modified to work with the changes from this PR, isn't it?

My suggestion is to create a new tag in `rrtmgp-data` whenever we need a new dataset for `rte-rrtmgp`. We then reference the dataset in `rte-rrtmgp` by the tag. This way, each version of `rte-rrtmgp` knows exactly what version of the data it is supposed to work with.

Alternatively, we can check the CMake documentation and fetch the data by a commit SHA. This way, we won't need so many tags in `rrtmgp-data` (we will have them only for the releases of `rte-rrtmgp`, I guess).

RobertPincus:
Good idea, thanks, see f368446.

PR #348, opened 2025-01-27
Relabel ice sizes as diameter, cloud optics by g-point or band, update API

RobertPincus:
Ice particles are labeled as diameter (#309). 

Cloud optical properties may be provided on bands or on g-points (#347)

C++ API to the cloud kernels are corrected (#345, #343)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,3 +38,14 @@ repos:
     rev: v7.0.0
     hooks:
       - id: pyproject-flake8
+  - repo: https://github.com/executablebooks/mdformat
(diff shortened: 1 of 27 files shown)


szy21:
Hi Robert, I wonder why the Pade approximant cloud optics were removed?

RobertPincus:
This is a question for @mjiacono

mjiacono:
It was not obvious how to develop the Pade approximant coefficients for the g-point optics, and we wanted to maintain consistency between the band and g-point cloud optics and to streamline future development.

szy21:
I see, thanks @mjiacono!

PR #350, opened 2025-02-03
Added to aways use cbool

makepath-alex:
Removed the option for using something else than cbool

### DIFF
--- a/rte-kernels/api/rte_types.h
+++ b/rte-kernels/api/rte_types.h
@@ -16,11 +16,7 @@ This header files C-compatible Boolean and floating point types (see mo_rte_type
 */
 #pragma once
 
-#ifdef RTE_USE_CBOOL
 using Bool = signed char;
-#else
-using Bool = int;
-#endif
 
 #ifdef RTE_USE_SP
 using Float = float;


RobertPincus (commented):
Is it clear that `Bool` is always `signed char`? Here's the [documentation](https://gcc.gnu.org/onlinedocs/gfortran/ISO_005fC_005fBINDING.html) for `gfortran`

RobertPincus:
@makepath-alex Can we not assume a C99 compiler and use native `bool`? 

https://stackoverflow.com/questions/1921539/using-boolean-values-in-c

makepath-alex:
Currently in mo_rte_kind we have
```
use, intrinsic :: iso_c_binding, only: c_float, c_double, c_long, c_int, c_bool
integer, parameter :: wl = c_bool
```

That makes Fortran compatible natively with C, using bool of size 1 byte as it is in C.

I think the current code is good enough, we just had to account for the change in pyRTE.

I am considering removing the C headers altogether as they are clumsy and very prone to errors.
I have an idea how to auto-generate them in the build process, so this week we might remove them completely.

Though for the sake of simplicity we will need to add rules for writing Fortran subroutines.
For example this would be hard to parse with a regex:

```
      real(wp),                           intent(in) :: press_ref_log_delta, &
                                                        temp_ref_min, temp_ref_delta, &
                                                        press_ref_trop_log
```
so we'll have to rewrite it to single line declarations
```
      real(wp),                           intent(in) :: press_ref_log_delta
      real(wp),                           intent(in) :: temp_ref_min, temp_ref_delta
      real(wp),                           intent(in) :: press_ref_trop_log
```

If you're ok with such kind of change, we can automate the process

skosukhin:
You can take parts of the dependency generator (the version in the [`autoconf`](https://github.com/earth-system-radiation/rte-rrtmgp/tree/autoconf/mkhelper) branch is a bit dated, the most recent version of the tool is [here](https://github.com/skosukhin/mkhelper/tree/develop/mkhelper)) to "preprocess" the Fortran files. For example, the following script produces lines that are easy to parse with regular expressions:
```python
import sys
import re

_re_line_continue_start = re.compile(r"^(.*)&\s*$")
_re_line_continue_end = re.compile(r"^\s*&")


def find_unquoted_string(string, line, quotes="'\""):
    skip = 0
    quote = None
    while 1:
        idx = line.find(string, skip)
        if idx < 0:
            return idx

        escaped = False
        for c in line[skip:idx]:
            if escaped:
                escaped = False
            elif c in quotes:
                if quote is None:
                    quote = c
                elif quote == c:
                    quote = None
            elif c == "\\" and quote:
                escaped = True

        if quote:
            skip = idx + len(string)
        else:
            return idx


def drop_comments_and_empty_lines(stream):
    for line in stream:
        comment_idx = find_unquoted_string("!", line)
        if comment_idx >= 0:
            line = line[:comment_idx] + "\n"
        if line and not line.isspace():
            yield line


def streamline_input(stream):
    stream = drop_comments_and_empty_lines(stream)
    for line in stream:
        # concatenate lines
        while 1:
            match = _re_line_continue_start.match(line)
            if not match:
                break

            next_line = next(stream, None)
            if next_line is None:
                break

            line = match.group(1) + re.sub(
                _re_line_continue_end, "", next_line
            )

        # split semicolons
        while 1:
            idx = find_unquoted_string(";", line)
            if idx < 0:
                break

            yield line[:idx] + "\n"
            line = line[idx + 1:]

        yield line


def main(inp):
    with open(inp) as f:
        for line in streamline_input(f):
            sys.stdout.write(line)


if __name__ == "__main__":
    main(sys.argv[1])
```

For example, instead of
```fortran
      !! temperature dimension in RRTMGP tables
    real(wp),                           intent(in) :: press_ref_log_delta, &
                                                      temp_ref_min, temp_ref_delta, &
                                                      press_ref_trop_log
      !! constants related to RRTMGP tables
    real(wp),    dimension(2,0:ngas,ntemp), intent(in) :: vmr_ref
```
you get
```fortran
    real(wp),                           intent(in) :: press_ref_log_delta,                                                       temp_ref_min, temp_ref_delta,                                                       press_ref_trop_log
    real(wp),    dimension(2,0:ngas,ntemp), intent(in) :: vmr_ref
```
which should be pretty easy to parse.

makepath-alex:
I played with a script to generate auto C headers with regex and it worked nicely, however a challenge was parsing multiline declarations such as this one:
```
      real(wp),                           intent(in) :: press_ref_log_delta, &
                                                        temp_ref_min, temp_ref_delta, &
                                                        press_ref_trop_log
```
As it was skipping the variable declarations on the second and third line and was simply returning 
```
      real(wp),                           intent(in) :: press_ref_log_delta, &
```
as in your example.

It could be fixed, but I don't see a good reason to complicate the parser, if we can agree to not write multiline declarations like the one above.

Though as you are the main contributors to the repo, I woudln't go ahead and make such a change without your approval.
So I am asking for your opinion

RobertPincus:
@makepath-alex I discussed with @skosukhin. We are happy to accept a PR that removes the C headers from the source,  includes a script to auto-generate the C-headers, and changes the Fortran API to accommodate those scripts. Could you do that?

RobertPincus (approved):

makepath-alex:
Absolutely. Will start a new PR for that

PR #355, opened 2025-02-10
Revive CI on Lumi

skosukhin:
This enables CI on LUMI with the Cray compiler and AMD GPUs again. The respective GitLab runner is registered at [gitlab.dkrz.de](https://gitlab.dkrz.de/icon-libraries/librte-rrtmgp), and we can have only one mirror now, which significantly simplifies maintenance.


RobertPincus:
@skosukhin Amazing, thanks so much. Still a draft because some of the tests are failing when the only change has been an increased size of some strings?

skosukhin:
For now, I just didn't want this work to get off the radar since the original #334 was automatically closed when you removed the `develop` branch.

RobertPincus:
Ah crap, sorry about that.

RobertPincus (approved):
Thanks @skosukhin for keeping the CI humming and also for the more general changes to the cloud optics.

Issue #358, opened 2025-02-20
Tests fail if `-fbounds-check` is not included in the compiler flags for gfortran


makepath-alex:
When building with CMake in **Release** mode, it **can** add the following flags to the currently set **FFLAGS**:
```sh
-O3 -Jmodules -fPIC -ffree-line-length-none -fpreprocessed
```
Note: For certain operations, all of these flags may be added, while for others, only `-O3` might be applied.

With GFortran, the RTE application is built and all tests pass when using the following flags:
```sh
-m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none
```
Interestingly, removing `-m64 -std=f2008` does not seem to affect the results, which is really odd given the information in #318.

However, without the `-fbounds-check` flag, test 17 (rte_sw_solver_unit_tests) fails with the following error:
```
test 17
      Start 17: rte_sw_solver_unit_tests

17: Test command: /data/vkm/code/makepath/rte-rrtmgp/build/tests/rte_sw_solver_unit_tests
17: Working Directory: /data/vkm/code/makepath/rte-rrtmgp/build/tests
17: Test timeout computed to be: 1500
17:  RTE SW solver unit tests
17:  Thin, scattering atmospheres
17:   mu0 = 1.00
17:  Direct flux doesn't match
17:    Net flux variants
17:    Subsetting invariance
17:    Vertical orientation invariance
17:    Linear in TOA flux
17:   mu0 = 0.50
17:  Direct flux doesn't match
17:    Net flux variants
17:    Subsetting invariance
17:    Vertical orientation invariance
17:    Linear in TOA flux
17:  RTE SW solver unit tests done
17: 
17: ERROR STOP 1
17: 
17: Error termination. Backtrace:
17: #0  0x7b1b15223960 in ???
17: #1  0x7b1b152244d9 in ???
17: #2  0x7b1b15225bb2 in ???
17: #3  0x6108b87f288c in ???
17: #4  0x6108b87ee27e in ???
17: #5  0x7b1b14e29d8f in __libc_start_call_main
17: 	at ../sysdeps/nptl/libc_start_call_main.h:58
17: #6  0x7b1b14e29e3f in __libc_start_main_impl
17: 	at ../csu/libc-start.c:392
17: #7  0x6108b87ee2b4 in ???
17: #8  0xffffffffffffffff in ???
17/23 Test #17: rte_sw_solver_unit_tests .........***Failed    0.14 sec
[...]

PR #359, opened 2025-02-20
Fix to optimization precision loss

makepath-alex:
#358

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -58,7 +58,7 @@ jobs:
             fpmodel: SP
     env:
       FC: gfortran-${{ matrix.gfortran-version }}
-      FFLAGS: "-m64 -std=f2008 -march=native -fmodule-private -fimplicit-none"
+      FFLAGS: "-m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan"
       CMAKE_BUILD_PARALLEL_LEVEL: 8
       VERBOSE:
       CTEST_PARALLEL_LEVEL: 8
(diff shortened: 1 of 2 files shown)


RobertPincus (commented):
Thanks @makepath-alex. I wonder if we want to make these changes to the CI? It seems like having debugging information should the tests fail would be useful. Would we rather add a separate set of testing with more release-like compiler flags?

.github/workflows/continuous-integration.yml
```diff
@@ -58,7 +58,7 @@ jobs:
             fpmodel: SP
     env:
       FC: gfortran-${{ matrix.gfortran-version }}
-      FFLAGS: "-m64 -std=f2008 -march=native -fbounds-check -fmodule-private -fimplicit-none -finit-real=nan"
```
skosukhin, line 61: These flags are not supposed to change the result and should stay. If they change the result, we should understand why (compiler bug, library but, etc.).

However, it feels right to drop `-march=native` (and probably `-m64`). It would be nice to ensure that dropping `-std=f2008` does not affect the results as well.

.github/workflows/continuous-integration.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       - name: Build libraries and tests
         run: |
           cmake -S . -B build -G "Ninja" \
-              -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+              -DCMAKE_BUILD_TYPE=Release \
```
skosukhin, line 119: Building Fortran code with `-O3` is a slippery slope in my experience. Alas, I can't tell more than that at the moment.

tests/rte_sw_solver_unit_tests.F90
```diff
@@ -126,7 +126,7 @@ program rte_sw_solver_unit_tests
     end if
     if(.not. allclose(sfc, &
                       toa_flux(:,1)*mu0_arr*exp(-sum(atmos%tau(:,:,1),dim=2)/mu0_arr), &
-                      tol=20._wp)) then ! Tolerances as big as 20 needed for GPU implementations
+                      tol=24._wp)) then ! Tolerances as big as 20 needed for GPU implementations
```
skosukhin, line 129:
```suggestion
                      tol=24._wp)) then ! Tolerances as big as 24 needed for GPU implementations
```

skosukhin (dismissed):

RobertPincus:
Closed with #360 which limits compiler flags

PR #363, opened 2025-02-25
Good old pure Ubuntu CI

skosukhin:


skosukhin:
> Can you explain why you use the `RelWithDebugInfo` CMake build type as opposed to `Debug`?

If I remember correctly, `Debug` means `-g` and `RelWithDebugInfo` means `-g -O2`. I would like to have tests with `-O2` and don't see a reason for running with just `-g`.

PR #368, opened 2025-03-14
Update version also in CMake. Remove dependency on DLB_EPSILON and FL…

makepath-alex:
The version in the CMAKELISTS.txt was not updated when we updated versions to 1.9.1

As a result the installed libraries are named with the wrong version:
```
librte.so -> librte.so.1.9.0
librte.so.1.9 -> librte.so.1.9.0
librte.so.1.9.0
```
This change will fix this problem. We'll have to remember for future releases to also update the CMake version.
I will add this to the README for releases.

The second part of this fix is I replaced the DBL_EPSILON and FLT_EPSILON with their respective constant values.
The reason is that currently they are used in the `rte_types.h` generation which is a header file we ship with the conda package
And now the header file looks like this:
```
const Float Float_epsilon = DBL_EPSILON;
``` 
Which requires the user of the header to bring a value for DBL_EPSILON or FLT_EPSILON which may cause issues.
Also for pyRTE to build we also need to define DBL_EPSILON but we don't use it...

For  a future release we may altogether remove this from the public header, because we don't need this to be public I think,


makepath-alex:
pre-commit.ci autofix

RobertPincus (approved):

PR #370, opened 2025-04-22
Change code structure to work around bug in ifx

tjahns:
The changes in this PR should not introduce any numerical differences but improve data access patterns in 3D interpolation slightly.


RobertPincus:
@peverwhee Can you please let @tjahns know if this solves your problem?

peverwhee:
Our intel (ifort) regression tests now run without issue. Thanks again!

RobertPincus:
@tjahns Thanks so much, I should have been more careful not to introduce a regression.

PR #379, opened 2026-01-07
Implement SSM optics

AndrewILWilliams:
This PR implements an idealized spectral representation of gas/cloud optics (termed 'SSM optics'), in the spirit of [Williams (2026)](https://arxiv.org/abs/2508.09353) for the longwave, and [Roemer et al. (2025)](https://www.nature.com/articles/s41612-025-01144-3) / [Cohen & Pincus (2025)](https://www.science.org/doi/full/10.1126/sciadv.adv6191) for the shortwave. 

### Overview
The basic idea is to represent the spectral structure of absorption coefficients using simplified analytical fits, a visual example for the longwave is provided below, where the raw absorption coefficients for $H_{2}O$ line absorption, $H_{2}O$ continuum absorption, and $CO_{2}$ line absorption are show in semi-transparent dots, and the simplified analytical fits (for $H_{2}O$ and $CO_{2}$ line absorption) are overlaid:

Such analytic fits reduce the empiricism of a correlated-k code by sparsely parameterizing the spectroscopy in terms of a handful of (physically-meaningful and easily interpretable) parameters. 

As we will show below the errors introduced by this approach are not negligible, and if the user requires a very accurate radiative transfer code, they should use RRTMGP. However, the SSM optics is sufficiently accurate to be used for many problems, particularly in the context of idealized climate modeling where a premium is placed on having parsimonious and _intelligible_ parameterizations.

###  Equations / Implementation

The SSM optics represents gaseous absorption ($\kappa$) as a series of overlapping `triangles`, each of the form: 

$$
\kappa = \kappa_{0} \exp\left(-\frac{|\nu - \nu_{0}|}{l}\right)
$$

where $l, \kappa_{0}, \nu_{0}$ are tunable spectroscopic parameters (1). The triangles are implemented in [mo_optics_ssm.F90](https://github.com/earth-system-radiation/rte-rrtmgp/blob/aee58c1e3ae283fdfd6766ebd681ee3fcc83e202/ssm/mo_optics_ssm.F90#L70). For the longwave and shortwave we have an array of shape `(ntriangles, 4)`  where the second dimension holds `[gas_index, kappa0, nu0, l]`. We currently have simplified representations of $H_{2}O$ and $CO_{2}$ in the longwave, and $H_{2}O$ in the shortwave (ozone is a todo).

For the shortwave, the user also defines a solar temperature and total solar irradiance, defaults are defined [here](https://github.com/earth-system-radiation/rte-rrtmgp/blob/aee58c1e3ae283fdfd6766ebd681ee3fcc83e202/ssm/mo_optics_ssm.F90#L38). 

For cloud optics, we define a single absorption coefficient ($\kappa$), single-scattering albedo ($\omega$), and asymmetry factor ($g$) for clouds in the longwave and shortwave; default values defined [here](https://github.com/earth-system-radiation/rte-rrtmgp/blob/aee58c1e3ae283fdfd6766ebd681ee3fcc83e202/ssm/mo_optics_ssm.F90#L45C1-L52C85). We do not currently differentiate between liquid and ice clouds, but this could be changed by altering [these lines](https://github.com/earth-system-radiation/rte-rrtmgp/blob/aee58c1e3ae283fdfd6766ebd681ee3fcc83e202/ssm/mo_optics_ssm.F90#L523) and defining a separate $\kappa, \omega, g$ for ice and liquid.

Optical depths and source functions (the Planck function) are calculated in `mo_optics_ssm_kernels.F90`. The optical depth calculation accounts for foreign pressure broadening of absorption coefficients if the `pref` input is non-zero. By default, the absorption coefficients are assumed to be evaluated at [pref = 500 hPa](https://github.com/earth-system-radiation/rte-rrtmgp/blob/aee58c1e3ae283fdfd6766ebd681ee3fcc83e202/ssm/mo_optics_ssm.F90#L101). 

#### rfmip-clear-sky test case

The figure below shows profiles of `rlu`,`rld`, `rsu`, `rsd` for a representative RFMIP site under present day conditions:

The figure below shows the correlation between broadband longwave/shortwave heating rates in the atmosphere (K/day) at 200 hPa, 500 hPa, and 800 hPa. Each dot represents a certain RFMIP site and experiment; I've discarded experiments which involve perturbations to species not represented in the SSM yet (e.g. CH4): 

There is a lack of SW heating at 200 hPa in the SSM due to the absence of ozone in the SW (I'll open an issue for this).

This figure scatters the net broadband TOA and SFC fluxes as predicted by RTE-SSM and RTE-RRTMGP:

The figure below shows histograms of the root mean squared error between RTE-SSM and RTE-RRTMGP (for the net broadband TOA and SFC fluxes) aggregated across all of the `expts` and `sites` in the rfmip-clear-sky test case. 

There are some key takeaways from these figures:
 - The SSM does a reasonable job of capturing the vertical structure of upwards/downwards fluxes in clear sky, and the heating rates. 
 - The SSM optics has biases compared to RRTMGP, and the accuracy is nowhere near what you might get out of a line-by-line model. This is unsurprising given the simplicity of the analytical fits using by the SSM.

Despite these errors, as demonstrated in [Williams (2026)](https://arxiv.org/abs/2508.09353), the SSM is still accurate enough to be useful for climate simulations.

#### all_sky test case

The `all_sky` test cases implements the idealized RCE profiles from the [RCEMIP](https://gmd.copernicus.org/articles/11/793/2018/gmd-11-793-2018.html) protocol, with some idealized clouds in there. The `all_sky` case is primarily there to act as an example for how to plumb the cloud optics, and not intended to comprehensively sample the covariability of clouds and meteorology.

The SSM's simplified representation of cloud optics nevertheless yields reasonable profiles of `rlu`,`rld`, `rsu`, `rsd` compared to RTE-RRTMGP:

### Footnotes:

(1) This functional form is motivated by the fact that absorption coefficients are organized into large-scale `bands` which appear as approximately straight lines when plotted on a semi-log plot (as above). For example, $CO_{2}$'s primary absorption feature basically looks like a triangle (two decaying exponentials) in the first plot. The same can also be said for $H_{2}O$, although the continuum regions are better described by flat lines (a 'gray' continuum). We currently ignore the continuum, but may end up approximating it as a triangle with a very large $l$, such that it looks like a flat line.

### DIFF
--- a/docs/build_documentation.sh
+++ b/docs/build_documentation.sh
@@ -1,11 +1,18 @@
 #!/bin/sh
 rm -rf ../../public
 mkdir ../../public
-ford ford_site/rrtmgp-fortran-interface.md
-ford ford_site/rrtmgp-kernels.md
-ford ford_site/rte-fortran-interface.md
-ford ford_site/rte-kernels.md
-ford ford_site/ssm.md
+#
+# Ford auto-documentation file
+#
+ford ford_templates/rrtmgp-fortran-interface.md
+ford ford_templates/rrtmgp-kernels.md
+ford ford_templates/rte-fortran-interface.md
+ford ford_templates/rte-kernels.md
+ford ford_templates/ssm.md
+
+#
+#
+#
 cd jekyll_site/
 bundle exec jekyll build
 cd _site/
--- a/docs/ford_templates/rrtmgp-kernels.md
+++ b/docs/ford_templates/rrtmgp-kernels.md
@@ -9,11 +9,9 @@ md_extensions: markdown.extensions.toc
 author: The RTE+RRTTMGP consortium
 github: https://github.com/earth-system-radiation/
 license: by
-src_dir: ../../rrtmgp/kernels
-exclude_dir:
-	../../rrtmgp/kernels/accel
-	../../rrtmgp/kernels/api
 output_dir: ../reference/rrtmgp-kernels
+src_dir: ../../rrtmgp/kernels
+exclude_dir: ../../rte-kernels/api ../../rte-kernels/accel
 ---
 
 These pages document the low-level computational kernels used by RRTMGP.
--- a/docs/ford_templates/rte-kernels.md
+++ b/docs/ford_templates/rte-kernels.md
@@ -9,11 +9,9 @@ md_extensions: markdown.extensions.toc
 author: The RTE+RRTTMGP consortium
 github: https://github.com/earth-system-radiation/
 license: by
-src_dir: ../../rte-kernels
-exclude_dir:
-	../../rte-kernels/accel
-	../../rte-kernels/api
 output_dir: ../reference/rte-kernels
+src_dir: ../../rte-kernels
+exclude_dir: ../../rte-kernels/api ../../rte-kernels/accel
 ---
 
 These pages document the low-level computational kernels used by RRTMGP.
--- a/docs/ford_templates/ssm.md
+++ b/docs/ford_templates/ssm.md
@@ -9,18 +9,15 @@ md_extensions: markdown.extensions.toc
 author: The RTE+RRTTMGP consortium
 github: https://github.com/earth-system-radiation/
 license: by
-src_dir: ../../ssm
-exclude_dir:
-	../../rte-kernels/accel
-	../../rte-kernels/api
 output_dir: ../reference/ssm
+src_dir: ../../ssm
 ---
 
 These pages document the Simple Spectral Models of gas and cloud optics used by RTE-SSM.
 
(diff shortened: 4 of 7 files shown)


RobertPincus (approved):
Thanks for this @AndrewILWilliams. A few changes outside of the addition of the SSM to note for our listeners at home:
- Code for reading scheme data (serially from netCDF files) is now in rrtmgp/data-loading-examples. This example is used by the examples. 
- The clear-sky and all-sky examples now run with the SSM in addition to RRTMGP (the optics scheme is chosen based on the same of the executable). Continuous integration exercises both examples with the SSM but doesn't verify the result. 
- Building the C-language headers is now off by default in CMake. 
- The version of the NAG compiler used in CI on Levante is updated to 7.2 (@skosukhin says this is incompatible with ICON but NAG 7.1 panics on the new example code) 

Note also that the SSM doesn't have OpenACC/OpenMP enabled but we hope to get to this. 

I'm approving the PR even though the Lumi CI isn't working because that host is unavailable for another few weeks. 

This is great, thanks for the contribution.

Issue #317, opened 2024-12-23
CMake refinements


RobertPincus:
Issues noted by @skosukhin outstanding from  #316

- [ ] Come up with a common prefix for custom CMake variables. For example, BUILD_TESTING is a standard variable and should keep the name but `KERNEL_MODE, FAILURE_THRESHOLD, RTE_ENABLE_SP`, etc. are project-specific and should have the same and a project-specific prefix. Usually, this would be the name of the project but `RTE_RRTMGP`_ seems to be rather long. Maybe we could use something like `RR_`.
The tests are currently missing dependencies on each other and, therefore, we cannot run them in parallel.
- [ ] : introduce the missing test `FIXTURES` and consider fetching the required data at the build time as one of the tests.
- [ ] : make it possible to install the libraries (and maybe the data files).
- [ ] : come up with the public aliases for the targets (e.g. `rte-rrtmgp:rte`, `rte-rrtmgp:rte-dp`, etc.).
- [ ] : make sure we can build rte-rrtmgp as a subproject (i.e. `add_subdirectory(rte-rrtmgp)` should work and expose the targets).
- [ ] : make sure that a superproject can find rte-rrtmgp (i.e. `find_project(rte-rrtmgp))` in the build directory, as well as in the installation prefix.
- [ ] : consider changing the directory layout. For example, it feels natural to move `rte-kernels` and `rte-frontend` to a new first-level subdirectory `rte`. Directories `gas-optics, rrtmgp-kernels and rrtmgp-frontend` could be moved to a new first-level subdirectory `rrtmgp`. 
- [ ] We also have a minor inconsistency: the kernel mode is named `extern` but the respective files are stored in the `api` subdirectory.
- [ ] : revisit .gitignore.
- [ ] :  extend formatting jobs to cover Python (isort, black, flake8), C/C++/CUDA (cmake-format) and Fortran (fprettify) source files.
- [ ] Remove setup.sh (unconventional and misleading) and provide the building instructions in README.md (or elsewhere)

RobertPincus:
A note on the directory layout: `gas_optics` belongs to `rte`; `rrtmgp` is one implementation

RobertPincus:
Two issues resolved in #319

RobertPincus:
Four issues resolved by #323

PR #381, opened 2026-01-15
Tidying up

RobertPincus:
- rte/ directory with frontend/ and kernels/ subdirectories
- Project-specific CMake macros all start with RTE_
- Omit `setup.sh`

Addresses #317


RobertPincus:
Also move `extensions/` -> `rte/extensions/`

PR #389, opened 2026-03-17
Update default triangle_params_def_lw values

AndrewILWilliams:
You only need two entries in triangle_params to fit H2O line absorption. Previously I was trying to fit the median-filtered H2O line absorption using three triangles, but that's unnecessary. I've re-run using a fit like: 

$$
\kappa = \kappa_{1} \exp\left(-\frac{|\nu|}{l_{1}}\right) + \kappa_{2} \exp\left(-\frac{|\nu - \nu_{2}|}{l_{2}}\right)
$$

where $\nu_{2}=1600 cm^{-1}$. 

The results from #379 are nearly identical, because you only need two triangles to fit this H2O line behavior (below)

image


AndrewILWilliams:
Thanks! My bad

Issue #390, opened 2026-03-24
rte/frontend/gas-optics-template/mo_gas_concentrations.F90 has non-conformant OpenACC clauses in it.


scamp-nvidia:
Hi, 

I work with NVIDIA on the NVHPC compiler. We are working on upstreaming OpenACC in LLVM as we work towards a next generation Fortran compiler with OpenACC support. This gives us an opportunity to fix some implementation problems that we've found in NVHPC's compilers conforming to the OpenACC standards. In testing, we identified that mo_gas_concentrations.F90 has non-conformant OpenACC in it that the current compiler incorrectly allows and looking at the current dev version of RRTMGP, we believe this still needs to be resolved. Notably, we believe there are issues with structures in that file like this (I simplified a bit to remove openmp and ifdef statements): 

```
    !$acc data copyout (array) present(this)
    if(size(this%concs(igas)%conc, 2) > 1) then
      !$acc kernels default(none) present(p)
      array(:) = this%concs(igas)%conc(1,:)
      !$acc end kernels
```

The new implementation will complain that 'igas' is not included in any data statements. The current NVHPC compiler allows this because it currently doesn't impose default rules to scalars, but the standard is very clear that it should be doing so. We want to ensure that when a new OpenACC compiler comes out, you are able to successfully run your code with it. In order to get this, we would like to suggest bringing this file into conformance. We believe you could just drop the 'default(none)' clause from those structures in that file or add an explicit data clause for 'igas'.

Please let me know if there are any questions or concerns!

RobertPincus:
Closed with #392

PR #392, opened 2026-03-25
Remove default(none) from OpenACC directives

RobertPincus:
Addresses #390


RobertPincus:
@scamp-nvidia Thanks for the heads up in #390. Does this seem to do the trick? It's the one place in the code base we used `default(none)` in OpenACC directives. 

I'm delighted that you were looking at this code!

scamp-nvidia:
Hey sorry for the delay - NVIDIA holiday. That looks like it should do it! That's the only non-conforming code we ran into with RRTMGP so far! If we run into anything else though, we'll let you know! Hopefully it will run with the new compiler right out of the box! :)

Issue #399, opened 2026-05-18
cloud_optics: is reice an effective radius or an effective diameter?


MHBalsmeier:
While using AI to debug and analyze an RTE+RRTMGP coupler I wrote, it pointed me to a naming inconsistency that got me confused. In all transparency, the following analysis was largely written by AI:

**`cloud_optics()`: is `reice` an effective radius or an effective diameter?**

I ran into a naming inconsistency I wanted to check my understanding of.

The internal storage in `rrtmgp/frontend/mo_cloud_optics_rrtmgp.F90` uses clearly diameter-named fields:

```fortran
real(wp) :: diamice_lwr = 0._wp, diamice_upr = 0._wp
```

However, the public accessor methods (same file) return those same fields under radius names:

```fortran
function get_min_radius_ice(this) result(r)
  r = this%diamice_lwr
end function get_min_radius_ice

function get_max_radius_ice(this) result(r)
  r = this%diamice_upr
end function get_max_radius_ice
```

And the bounds check inside `cloud_optics()` validates `reice` — documented as "ice particle effective radius" — against `this%diamice_lwr`/`this%diamice_upr`.

The all-sky example (`examples/all-sky/rrtmgp_allsky.F90`) adds further evidence:

```fortran
rei_val = 0.5 * (cloud_optics%get_min_radius_ice() + cloud_optics%get_max_radius_ice())
```

This yields `rei_val = 95 µm`, which corresponds to the midpoint of the Yang et al. (2013) $D_{ge}$ range [10, 180] µm — not any physically plausible ice effective radius.

Furthermore, the ncdump output of the coefficients files is also confusing to me:

```
	float diamice_lwr ;
		diamice_lwr:standard_name = "radice_lwr" ;
		diamice_lwr:description = "Ice particle size lower bound for g-point LUT interpolation" ;
		diamice_lwr:units = "microns" ;
...
data:

 diamice_lwr = 10 ;

 diamice_upr = 180 ;
}
```

Is the conclusion correct that `reice` (and `rei` in the all-sky example) should in fact be supplied as the generalized effective **diameter** $D_{ge}$ in µm, and that the "radius" naming in the public API is simply a historical artefact that was never corrected?

Chiil:
This issue relates to: https://github.com/earth-system-radiation/rte-rrtmgp/pull/342

RobertPincus:
[...]

PR #401, opened 2026-05-25
Effective radius vs. diameter naming fix

MHBalsmeier:
In mo_cloud_optics_rrtmgp, reice has been renamed to diamice, since for ice, the effective diameter is used. Furthermore, one comment and one write statement have also been updated accordingly.

This refers to #399 .

For now, I did not change the names of the functions get_min_radius_ice and get_max_radius_ice, since this might break users' couplers. Do you want me to rename them as well?


RobertPincus:
@MHBalsmeier Thanks very much for taking this initiative. 

- It might be nicer to use `dgice` or `dgeice` to denote the "generalized effective diameter" and stay consistent with the `re` notation for liquid. 
- Please do go ahead and rename `get_min_radius_ice()` and `get_max_radius_ice()`. It might break user workflows, as you say; it might also alert people who haven't realized the sizes are diameters.

MHBalsmeier:
@RobertPincus Alright, let's see what it looks like now.

RobertPincus (approved):
Thanks for taking care of this detail - it will reduce confusion about what is meant.

PR #403, opened 2026-06-23
Gas optic utilities reorganization

RobertPincus:
Consolidates some low-level procedures used by gas optics routines. In kernels/ because the procedures don't use RTE's derived types. 

There's an inconsistency in that `get_layer_mass` gets the mass by species ordered as `(gas, col, layer)` while `get_layer_number()` returns only the number density (per cm^-2) of dry air molecules. It would make sense to have the latter return the number of each species but RRTMGP wants an order `(col, layer, gas)`...


RobertPincus:
Removing windows-2022 from the conda-compatible builds since a) the current conda recipe doesn't provide Windows; and b) the netcdf-fortran conda module now seems to be flang and not gfortran.

PR #404, opened 2026-07-01
Various fixes and workaround for amdflang

dindon-sournois:
Some fixes to be able to compile with latest AMD's flang with OpenMP offload
- using parameter arrays in kernel seems unsupported, use global module variable instead
- some array assignment statement in **serial** `omp target` directives would cause compiler link errors
  + fixed in ROCm 7.13 beta
  + replaced by explicit parallel loops, as `workdistribute` isn't supported yet

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,8 @@ if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU AND NOT RTE_RRTMGP_Fortran_is_NEC)
 endif()
 
 if(${CMAKE_Fortran_COMPILER_ID} MATCHES LLVMFlang)
+  # replace parameter local array in kernel by module global array use explicit
+  # openmp target loops
   add_compile_definitions(AMDFLANG_WORKAROUND)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,10 @@ if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU AND NOT RTE_RRTMGP_Fortran_is_NEC)
   add_compile_options($<$<COMPILE_LANGUAGE:Fortran>:-ffree-line-length-none>)
 endif()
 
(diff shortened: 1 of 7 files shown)


RobertPincus:
pre-commit.ci autofix

RobertPincus:
@dindon-sournois Thanks for keeping the codes working with OpenMP GPU offload. Please fix the errors in the preprocessor directives. 

I have no way to include OpenMP GPU my continuous integration. If you have ideas for doing so I would love to hear them

RobertPincus:
@dindon-sournois Sorry, I guess the directive bugs were introduced by pre-commit, but it would be ideal if pre-commit was able to run on the entire code base

dindon-sournois:
For CI if you already have a nvfortran OpenACC test you could just replace in the compilation options `-acc` by `-mp`, the fixes I've added for amdflang should also work for the Nvidia compiler if they are enabled in the CMakelists

For the pre-commit issue, do you know how would I test it?

RobertPincus:
If you have `pre-commit` in your environment use `pre-commit install` and the checks will run locally before each commit. Then it's not possible to make a commit that doesn't follow the rules.

dindon-sournois:
@RobertPincus I've rebased my code based on pre-commit fixes and also fixed the `#ifdef` mismatches, not sure why I didn't catch that

RobertPincus (approved):
Thanks for the help.

It's great if these are removed if/when the AMD compiler catches up (or if the code isn't following a standard we can make it do so).

dindon-sournois:
For the use of `parameter` array in a kernel I'd say it's something that will be fixed (AMD now have a reproducer of this issue).
As for the rewriting of loops, I think the current version is preferable for performance while `target teams workdistribute` (with is more or less the equivalent of `acc kernels`) is still unsupported. To my knowledge it is both unsupported by AMD and Nvidia.

RobertPincus:
@dindon-sournois If you have access to Lumi please feel free to propose more configurations to the Gitlab CI

PR #407, opened 2026-07-09
amdflang workaround for using mod with real8 values and array operands inside OpenMP kernel

dindon-sournois:


dindon-sournois:
This bug looks similar to a bug I encountered when using `omp target` serial kernels on operations with array operands, but this one isn't fixed with the next ROCm release (7.13 beta / 8.0)

RobertPincus (approved):
An increasing number of `AMDFLANG_WORKAROUND` is kind of ugly... as the compiler matures it would be great to get rid of these as possible.

PR #408, opened 2026-07-10
Add CTest configuration for rte_examples

pradiptasam:
Adds CMake/CTest support for the rte_examples executable in examples/rte-examples/. The atmospheric state data (RFMIP, CKDMIP, RCE) is fetched from the rte-examples repository as a CTest fixture using the same ExternalProject pattern used for rrtmgp-data. An alternative path can be supplied via the RTE_EXAMPLES_DATA cache variable to skip the fetch.

Twelve tests are registered covering all combinations of optics scheme (RRTMGP, SSM), spectral band (LW, SW), and atmospheric state (RFMIP, CKDMIP, RCE). All tests depend on the fetch_rrtmgp_data and fetch_rte_examples_data fixtures, ensuring data is in place before any run is attempted.

### DIFF
--- a/examples/rte-examples/CMakeLists.txt
+++ b/examples/rte-examples/CMakeLists.txt
@@ -15,6 +15,54 @@ target_link_libraries(rte_examples_io PRIVATE shared-utils)
 
 # ##############################################################################
 #
+# Fetch the rte-examples atmospheric state data
+#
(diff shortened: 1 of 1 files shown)


RobertPincus (approved):
Beautiful, thanks so much. I see that the Intel compilers are complaining but we'll fix that separately.

PR #409, opened 2026-07-13
Fix ifx compatibility: slice varnames to avoid blank entries in pack

pradiptasam:
When compiled with ifx, pack(varnames, mask=is_gas(1:ivar)) over the full max_vars-sized array passes uninitialized blank entries to available_gases%init(), which then (correctly) rejects them as duplicate empty gas names. Other compilers happen not to trigger this because their uninitialized slots don't survive the mask check.

Fix: count the actual variables found in the NetCDF file (ngases) and slice varnames(1:ngases) before calling pack.


RobertPincus (approved):
Thanks @pradiptasam you're excellent, that's a terrific catch.

PR #412, opened 2026-07-15
Add rte-examples to CI

RobertPincus:
Having added solutions from the `rte-examples` problem to the `rrtmgp-data` repo and created a (very small!) `ssm-data` repo, this PR adds tests comparing solutions to those reference values.

### DIFF
--- a/ssm/CMakeLists.txt
+++ b/ssm/CMakeLists.txt
@@ -25,3 +25,50 @@ set_target_properties(
 )
 
 install(TARGETS ssm EXPORT rte-rrtmgp-targets)
+
+# ##############################################################################
+#
+# Fetch the rte-examples atmospheric state data
+#
+if(SSM_DATA)
+  add_test(NAME fetch_ssm_data COMMAND ${CMAKE_COMMAND} -E true)
+  message(
+    NOTICE
+    "Using an external dataset from ${SSM_DATA}: the data files will not be installed"
+  )
+else()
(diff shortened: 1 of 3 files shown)


RobertPincus:
@pradiptasam Any idea why this CMake setup is failing on the CI runners? It's working fine on my local host... 

Also if I'm being overly verbose with the CMake please let me know

ssm/CMakeLists.txt
```diff
@@ -25,3 +25,50 @@ set_target_properties(
...
+  install(
+    FILES # cmake-format: sort
+          ${SSM_DATA}/ssm-lw-ckdmip.nc
+          ${SSM_DATA}/ssm-lw-rce.nc
+          ${SSM_DATA}/ssm-lw-rfmip.nc
+          ${SSM_DATA}/ssm-sw-ckdmip.nc
+          ${SSM_DATA}/ssm-sw-rce.nc
+          ${SSM_DATA}/ssm-sw-rfmip.nc
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/rte-rrtmgp/
+  )
```
pradiptasam: These are not needed unless and until the test `fetch_ssm_data` is run, and I checked that it is not currently being run.  The `make install` step looks for these files, but they are not generated until the test is run. Hence the error.  We can comment out these lines now, or delete and bring them later.

PR #413, opened 2026-07-21
Fetch scheme/test/reference data via release tarballs at build time

pradiptasam:
## Summary

Reworks how the RRTMGP, SSM, and rte-examples data are acquired, replacing
git clones triggered by ctest with release-tarball downloads that happen at
build time.

## Motivation

Previously, each dataset was fetched by cloning its git repo through an
`ExternalProject_Add(GIT_REPOSITORY ...)` marked `EXCLUDE_FROM_ALL`, and the
download was triggered indirectly by a ctest `fetch_*_data` test. This had
several drawbacks:

- Downloading data as a side effect of "running a test" conflated test execution with data acquisition.
- Fetching required `git` and git-protocol access at build time.
- The install step copied reference/test data that downstream users of the library don't need.

## Changes

For each of the three datasets (`rrtmgp-data`, `ssm-data`, `rte-examples`):

- Switch `ExternalProject_Add` from `GIT_REPOSITORY`/`GIT_TAG` to a `URL`
  pointing at the release tarball:
  - rrtmgp-data → v1.9.1
  - ssm-data → v0.0.1
  - rte-examples → v0.1.1
- Remove `EXCLUDE_FROM_ALL` so the data target is part of the default `ALL` target and is fetched during `cmake --build`.
- Drop the `install(CODE "execute_process(... --target ...)")` re-trigger; the files already exist in the build tree at install time, so `install(FILES ...)` copies them directly.
- Replace the `fetch_*_data` test (which used to trigger the download) with a presence-check fixture that verifies the required `.nc` files exist and fails with an actionable "build the project first" message otherwise. The  `FIXTURES_SETUP`/`FIXTURES_REQUIRED` wiring is preserved so tests still run in the correct order.

New helper script: `cmake/CheckData.cmake` which works for all the three data files. 

**Install scope:**

- Only scheme data (the RRTMGP `.nc` coefficient/cloud/aerosol files) is installed via `install(FILES ...)`.
- Reference data (`ssm/.../reference/ssm-*.nc`) and test-input data (`rte-examples/*-states.nc`) are still fetched for the test suite but are no longer installed, since they are not runtime dependencies of the library.


RobertPincus:
@pradiptasam Thanks for these changes. To clarify: all data including scheme and testing data are downloaded at build time, but only the scheme data are "installed"?

pradiptasam:
> @pradiptasam Thanks for these changes. To clarify: all data, including the scheme and testing data, are downloaded at build time, but only the scheme data are "installed"?

Yes, that's what it should do. All data are downloaded at build time, but only the scheme data are copied during installation. I have the same changes applied to all the data files to download them at build time, and I deleted only the reference and test data being copied. You have to consider one thing, though: you still need to run cmake with `-DRTE_BUILD_TESTING=ON` to get the data files downloaded. I will move them outside the loop in an upcoming MR, as doing that in this MR would make the changes unreadable. After that, you would not need to use `-DRTE_BUILD_TESTING=ON` anymore to get the data files.

RobertPincus:
> > @pradiptasam Thanks for these changes. To clarify: all data, including the scheme and testing data, are downloaded at build time, but only the scheme data are "installed"?
> 
> Yes, that's what it should do. All data are downloaded at build time, but only the scheme data are copied during installation. I have the same changes applied to all the data files to download them at build time, and I deleted only the reference and test data being copied. You have to consider one thing, though: you still need to run cmake with `-DRTE_BUILD_TESTING=ON` to get the data files downloaded. I will move them outside the loop in an upcoming MR, as doing that in this MR would make the changes unreadable. After that, you would not need to use `-DRTE_BUILD_TESTING=ON` anymore to get the data files.

Great, thanks for the explanation.

RobertPincus (approved):
The Gitlab CI is failing because one of our hosts was down.

PR #414, opened 2026-07-21
Move data download out of RTE_BUILD_TESTING guard and quiet CMake output

pradiptasam:
## Summary

Follow-up refinements to the data-fetching setup. Three focused changes:

1. Opt into CMake policy `CMP0135` to silence the URL-download timestamp warning.
2. Reduce build-time console noise from the archive download/extract step.
3. Move the RRTMGP and SSM data fetch/install out of the `RTE_BUILD_TESTING` guard and into their own subdirectories, so scheme data is acquired and installed regardless of whether tests are enabled.

### Silence CMP0135 warning
URL-based `ExternalProject_Add` downloads emit a `CMP0135` developer warning on CMake >= 3.24. Set the policy to `NEW` (extraction-time timestamps) once at the top of the root `CMakeLists.txt`, guarded by `if(POLICY CMP0135)` so it is a no-op on older CMake (project minimum is still 3.18).

### Quiet the build output
Add `LOG_DOWNLOAD TRUE` to the `ExternalProject_Add` calls so that the verbose download/extract status is written to a log file rather than the console (CMake still prints the log automatically on failure).

### Move data fetch outside the testing guard
Previously, the `rrtmgp-data` fetch/install lived in the root `CMakeLists.txt` inside `if(RTE_BUILD_TESTING)`, meaning scheme data was only fetched/installed when tests were enabled. Scheme data is a runtime dependency of the library, so:

- Move the `rrtmgp-data` block into `rrtmgp/CMakeLists.txt`.
- Move the `ssm-data` block out of its `RTE_BUILD_TESTING` guard in `ssm/CMakeLists.txt`.
- Propagate `RRTMGP_DATA` / `SSM_DATA` to the parent scope via `PARENT_SCOPE` so sibling subdirectories (`examples`, `tests`) inherit the resolved value.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,57 +102,6 @@ exit(tuple(map(int, xarray.__version__.split('.'))) < (0, 12, 2))"
 
   find_package(NetCDF_Fortran REQUIRED)
 
-  if(RRTMGP_DATA)
-    add_test(NAME fetch_rrtmgp_data COMMAND ${CMAKE_COMMAND} -E true)
-    message(
-      NOTICE
-      "Using an external dataset from ${RRTMGP_DATA}: the data files will not be installed"
-    )
-  else()
-    set(RRTMGP_DATA "${PROJECT_BINARY_DIR}/rrtmgp-data")
-
-    include(ExternalProject)
-    # Fetched at build time (part of the default ALL target) so the data is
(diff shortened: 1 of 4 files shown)


RobertPincus (approved):
So nice, thanks a lot.

